PR #25, opened 2024-11-13
Single-task POYO

vinamarora8:
Tested on entire MP dataset for up to 60 epochs. Works as expected. On 4xA40s: time per epoch: ~40s, time per validation: ~10s. I believe we are ready for training POYO-MP.

There is some ugliness wrt `subtask_index` handling. But I am deliberately leaving that until brainsets merges the interval-based approach so we don't have to re-engineer it later.


examples/poyo/configs/base.yaml
```diff
@@ -0,0 +1,48 @@
...
+num_workers: 4
+seed: 42
+
+subtask_weights:
+  - 1.0  # RANDOM
+  - 0.1  # HOLD
+  - 5.0  # REACH
+  - 1.0  # RETURN
+  - 0.1  # INVALID
+  - 0.0  # OUTLIER
```
vinamarora8, line 29: This is hacky. Will clean once the interval-based movement phases are merged into brainsets.

vinamarora8:
@mazabou Ready for review. If possible, we should merge this in before adding `examples/poyo/finetune.py`

mazabou (commented):
Looks good overall. I left a few comments. 

I made the update to `brainsets`, I can attempt to incorporate them here.

examples/poyo/README.md
```diff
@@ -0,0 +1,30 @@
+# POYO 🧠
+Official codebase for POYO published at NeurIPS 2023
+[[project page]](https://poyo-brain.github.io/)
+[[arxiv]](https://arxiv.org/abs/2310.16046)
+
+### Training
+To train POYO-MP you can run:
+```bash
+python train.py --config-name train_poyo_mp.yaml
+```
```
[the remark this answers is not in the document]
vinamarora8: Yes, we should remember to do that

examples/poyo/configs/base.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+data_root: ./data/processed/
+log_dir: ./logs
+
+sequence_length: 1.0  # in seconds
+latent_step: 0.125  # in seconds
+modality_name: cursor_velocity_2d
+
+epochs: 1000  # if -1, will use steps
+steps: -1  # if -1, will use epochs. Epochs take precedence.
```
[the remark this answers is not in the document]
vinamarora8: oh, I just copied these from poyo_plus I think. Although I added them there. Not needed for supervised cases. I'll remove it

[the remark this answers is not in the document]
mazabou: yes, it's not you who added it. I meant to ask this for a while. I know we had it before.

torch_brain/utils/stitcher.py
```diff
@@ -71,7 +74,159 @@ def stitch(timestamps: torch.Tensor, values: torch.Tensor) -> torch.Tensor:
     return averages
 
 
-class StitchEvaluator(L.Callback):
+class MultiSessionDecodingStitchEvaluator(L.Callback):
```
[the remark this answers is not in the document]
vinamarora8: It is specific to multi-session situations because it requires the tokenizer to add `"session_id"` field to the batch-data. In single session situations users shouldn't be expected to add this field.

I believe having the `Stitcher` class (independent of Lightning, the idea we discussed on slack) would end up making this situation more unified. This would take some thinking and experimenting; which I can probably get to next week, but this week I'm tight on time. Until then, I feel this is a good solution.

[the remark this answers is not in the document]
mazabou: I agree with the second point. 

Re: `"session_id"`, I also agree. My solution is that instead of rewritting the whole class, we just add an argument to the existing one that would be "single_session=True". And it will not expect to get "session_id" in that case. everything else will be exactly the same.

torch_brain/utils/stitcher.py
```diff
@@ -71,7 +74,159 @@ def stitch(timestamps: torch.Tensor, values: torch.Tensor) -> torch.Tensor:
...
+        self.on_validation_epoch_start(*args, **kwargs)
+
+    def on_test_batch_end(self, *args, **kwargs):
+        self.on_validation_batch_end(*args, **kwargs)
+
+    def on_test_epoch_end(self, *args, **kwargs):
+        self.on_validation_epoch_end(*args, **kwargs, prefix="test")
+
+
+class MultiSessionMultiTaskStitchEvaluator(L.Callback):
```
vinamarora8, line 229: Note to self: change the name to `MultiSessionMultiTaskDecodingStitchEvaluator`

torch_brain/utils/stitcher.py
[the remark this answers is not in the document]
vinamarora8: Removed "MultiSession" from the name

torch_brain/models/poyo.py
```diff
@@ -0,0 +1,354 @@
...
+            output_weights = np.ones(len(output_values), dtype=np.float32)
+        else:
+            output_weights = self.subtask_weights[output_subtask_index]
+
+        # Mask for the output predictions
+        output_mask = np.ones(len(output_values), dtype=bool)
+        if self.eval:
+            # During eval, only evaluate on the subtask specified in the config
+            target_subtask_index = data.config["eval_subtask_index"]
+            output_mask = output_mask & (output_subtask_index == target_subtask_index)
```
mazabou, line 329: this will be replaced so I will not review it

torch_brain/models/poyo.py
```diff
@@ -0,0 +1,354 @@
...
+        subtask_weights (List[float]): Loss-weights for different subtasks.
+    """
+
+    def __init__(
+        self,
+        unit_tokenizer: Callable,
+        session_tokenizer: Callable,
+        latent_step: float,
+        num_latents_per_step: int,
+        modality_spec: ModalitySpec,
```
mazabou, line 253: should we name it `readout_spec`? `modality_spec` is more consistent, but could be confusing since spikes itself is a modality. and in the future we can change the input modality. so for consistency I would suggest calling this `readout_spec`

torch_brain/models/poyo.py
```diff
@@ -0,0 +1,354 @@
...
+            # latent sequence
+            "latent_index": latent_index,
+            "latent_timestamps": latent_timestamps,
+            # output sequence
+            "output_session_index": pad8(session_index),
+            "output_timestamps": pad8(output_timestamps),
+            "output_mask": pad8(output_mask),
+            # ground truth targets
+            "target_values": pad8(output_values),
+            "target_weights": pad8(output_weights),
```
mazabou, line 346: here padding is used instead of chaining. how are you dealing with the loss when it comes to padding values?

mazabou (commented):
Review of `torch_brain/models`

examples/poyo/train.py
```diff
@@ -0,0 +1,362 @@
...
+from torch_brain.transforms import Compose
+
+# higher speed on machines with tensor cores
+torch.set_float32_matmul_precision("medium")
+
+logger = logging.getLogger(__name__)
+
+
+@hydra.main(version_base="1.3", config_path="./configs", config_name="train.yaml")
+def main(cfg: DictConfig):
```
mazabou, line 38: the python convention is that `main()` goes at the bottom of the script. so this should be the last function.

examples/poyo/train.py
```diff
@@ -0,0 +1,362 @@
...
+        target_weights = target_weights[output_mask]
+
+        if self.modality_spec.loss_fn == "mse":
+            loss = torch.nn.functional.mse_loss(
+                output_values, target_values, reduction="none"
+            )
+            loss = loss * target_weights[:, None]
+            loss = loss.mean()
+        else:
+            raise NotImplementedError("Only MSE loss is supported for now.")
```
mazabou, line 190: why are we going backwards here?

torch_brain/utils/callbacks.py
```diff
@@ -20,7 +20,7 @@ def on_train_epoch_start(self, trainer, pl_module):
     def on_train_epoch_end(self, trainer, pl_module):
         if self.enable:
             epoch_time = time.time() - self.start_time
-            pl_module.log("epoch_time", epoch_time)
+            pl_module.log("epoch_time", epoch_time, sync_dist=True)
```
mazabou, line 23: why do we need to sync_dist here? isn't everything blocked until the last gpu is done? also, why does this need to be accurate?

torch_brain/utils/stitcher.py
```diff
@@ -71,7 +74,159 @@ def stitch(timestamps: torch.Tensor, values: torch.Tensor) -> torch.Tensor:
...
+    timestamps for each session. This cache is updated at the end of each batch.
+    Finally, at the end of the epoch, the cache is coalesced and the metric is
+    computed for each session.
+    """
+
+    def __init__(
+        self,
+        session_ids: Iterable[str],
+        modality_spec: Optional[ModalitySpec] = None,
+        metric_factory: Optional[Callable[[int], ModalitySpec]] = None,
```
mazabou, line 118: what's `metric_factory`? this nullifies the option to specify a different metric per dataset?

mazabou: are we creating a concept that will not work for POYOPlus?

examples/poyo/configs/train_poyo_mp.yaml
```diff
@@ -0,0 +1,15 @@
...
+
+train_transforms:
+  - _target_: torch_brain.transforms.UnitDropout
+    max_units: 1000
+    min_units: 60
+    mode_units: 300
+    peak: 4
+
+wandb:
+  run_name: poyo_mp_mini
```
mazabou, line 15: why mini?

mazabou (commented):
Review part 2 of 2

torch_brain/models/poyo.py
vinamarora8: We should change this in the spec python file as well in that case. I just went with whatever name was used there. I support renaming it to readout_spec at both places.

examples/poyo/train.py
vinamarora8: yeah we can rename this to `train()` or smth. Otherwise I don't like having both DataModule and TrainWrapper at the top. Too much code before you get to what this script is actually running. Happy to reconsider.

examples/poyo/train.py
vinamarora8: Just wanted to get it done without having to test _this_ code on datasets with other kinds of losses. We can bring back other losses when we've tested them. Imo we shouldn't include a feature we haven't tested. Adding more losses can be separate PRs.

torch_brain/utils/callbacks.py
vinamarora8: It's not needed in theory. Lightning complains if we don't have sync_dist=True. I'd rather not see that warning. This incurs negligible cost.

torch_brain/utils/stitcher.py
vinamarora8: `metric_factory` is when you wanna use a custom metric bypassing the modality_spec framework.

vinamarora8: This will not work for POYOPlus. This stitcher is not meant for handling multi-task output

vinamarora8: I think I have made a mistake here. metric_factor's type should be `Optional[Callable[[int], torchmetrics.Metric]`

examples/poyo/configs/train_poyo_mp.yaml
vinamarora8: Idk, ill call it poyo_mp

torch_brain/models/poyo.py
```diff
@@ -0,0 +1,354 @@
...
+    recordings.
+
+    TODO: Document the model architecture
+
+    Args:
+        dim: Hidden dimension of the model
+        dim_out: Dimension of the decoding targets
+        dim_head: Dimension of each attention head
+        num_latents: Number of unique latent tokens (repeated at every latent step)
+        depth: Number of processing layers (self-attentions in the latent space)
```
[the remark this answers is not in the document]
vinamarora8: Sure, it can be a separate PR focused on POYOPlus edits.

torch_brain/models/poyo.py
```diff
@@ -0,0 +1,354 @@
...
+    <https://arxiv.org/abs/2310.16046>`.
+
+    POYO is a transformer-based model for neural decoding from electrophysiological
+    recordings.
+
+    TODO: Document the model architecture
+
+    Args:
+        dim: Hidden dimension of the model
+        dim_out: Dimension of the decoding targets
```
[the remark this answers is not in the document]
vinamarora8: done

torch_brain/models/__init__.py
```diff
@@ -1 +1,2 @@
 from .poyo_plus import POYOPlus, POYOPlusTokenizer
+from .poyo import POYO, poyo_mp, POYOTokenizer
```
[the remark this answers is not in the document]
vinamarora8: done

vinamarora8:
Note I have removed `strategy=ddp_find_unused_parameters_true ...` and `accelerator = ...` arguments of lightning trainer because we don't need them for this code.

torch_brain/models/poyo.py
```diff
@@ -0,0 +1,354 @@
...
+        input_token_type: TensorType["batch", "n_in", int],
+        input_mask: Optional[TensorType["batch", "n_in", bool]] = None,
+        # latent sequence
+        latent_index: TensorType["batch", "n_latent", int],
+        latent_timestamps: TensorType["batch", "n_latent", float],
+        # output sequence
+        output_session_index: TensorType["batch", "n_out", int],
+        output_timestamps: TensorType["batch", "n_out", float],
+    ) -> TensorType["batch", "n_out", "dim_out", float]:
+        """Forward pass of the POYO+ model.
```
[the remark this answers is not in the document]
mazabou: To clarify, I am not referring to chain vs. pad. What I am suggesting is replicating the `unpack_output=True` behavior, so that instead of returning `B x T_out x D_out`, we return `List[*, D_out]` of length `B`. The logic for this is super simple. The typical user is not aware that masking is being used! — it's happening in the tokenizer/collate. so they wouldn't know that the output they are getting will need to be "unpadded", in which case they would have to know how to access to the `output_mask` inside of the batch dict. This would of course happen rarely but it is likely incase some dataset has missing values etc...

My suggestion is to include a single line of code:
```
if unpack_output:
     return [output[b][output_mask[b]] for b in range(output.size(0))]
```

torch_brain/models/poyo.py
mazabou: what do you mean by the spec python file? you mean renaming `ModalitySpec` to `ReadoutSpec`?

examples/poyo/train.py
mazabou: let's put it back the way it was please. if you want to make the change across the board, it should be a separate PR/ a separate discussion.

examples/poyo/train.py
mazabou: we have `compute_loss_or_metric` that we can use no?

torch_brain/utils/stitcher.py
mazabou: this is a new concept. I would rather it be in a separate PR/discussion. For now, I do not see why are introducing two completely different ways for using the stitcher, based on whether it's POYO or POYOPlus. It will be extremely confusing.

vinamarora8: is your issue with the existence of `metric_factory` as an optional parameter?

examples/poyo/train.py
vinamarora8: forgot about compute_loss_or_metric. fixed.

torch_brain/models/poyo.py
vinamarora8: Yes

torch_brain/models/poyo.py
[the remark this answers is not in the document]
vinamarora8: Fair point. Done.

torch_brain/utils/stitcher.py
vinamarora8: removed

torch_brain/models/poyo.py
vinamarora8: Renamed `modality_spec` to `readout_spec`. Done.

mazabou (approved):
Look great! Thank you for the colossal amount of work. Good to merge.

Just checking, you are going to merge this into main correct?

PR #30, opened 2024-11-27
Add eval_interval field and normalization fields to POYO

vinamarora8:
This has a very high loss value on latest `brainsets`. Idk what the problem could be.


torch_brain/models/poyo.py
```diff
@@ -349,6 +357,30 @@ def __call__(self, data: Data) -> Dict:
...
+        if len(interval) == 0:
+            return np.zeros_like(x, dtype=bool)
+
+        x_expanded = x[:, None]
+        y = np.logical_and(
+            x_expanded >= interval.start[None, :], x_expanded < interval.end[None, :]
+        )
+        y = np.logical_or.reduce(y, axis=1)
+        assert y.shape == x.shape
+        return y
```
vinamarora8, line 386: This should eventually be moved to be a method of `temporaldata.Interval`.

examples/poyo/configs/base.yaml
```diff
@@ -16,12 +16,10 @@ num_workers: 4
...
-  - 1.0  # RANDOM
-  - 0.1  # HOLD
-  - 5.0  # REACH
-  - 1.0  # RETURN
-  - 0.1  # INVALID
-  - 0.0  # OUTLIER
+  random_period: 1.0
+  hold_period: 0.1
+  reach_period: 5.0
+  return_period: 1.0
```
vinamarora8, line 23: Note: We're missing `invalid_period`.

vinamarora8: I am seeing a difference in POYO-MP performance curves because of this. We should put back Invalid phase and test before merging.

PR #31, opened 2024-11-27
Resolve weights based on interval membership

mazabou:
This PR simplifies the interface used to determine weights based on whether a point belongs to a given interval.

### DIFF
--- a/examples/poyo/configs/dataset/perich_miller_population_2018.yaml
+++ b/examples/poyo/configs/dataset/perich_miller_population_2018.yaml
@@ -81,6 +81,11 @@
       - m_20150625_center_out_reaching
       - m_20150626_center_out_reaching
   config:
+    readout_id: cursor_velocity_2d
+    metrics:
+      - metric:
+          _target_: torchmetrics.R2Score
+          num_outputs: 2
     weights:
       movement_periods.random_period: 1.0
       movement_periods.hold_period: 0.1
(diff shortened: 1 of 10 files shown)


examples/poyo/configs/dataset/perich_miller_population_2018.yaml
```diff
@@ -110,4 +116,10 @@
     - m_20140221_random_target_reaching
     - m_20140224_random_target_reaching
   config:
+    weights:
+      movement_periods.random_period: 1.0
+      movement_periods.hold_period: 0.1
+      movement_periods.reach_period: 5.0
+      movement_periods.return_period: 1.0
+      cursor_outlier_segments: 0.0
```
vinamarora8, line 124: Does this work? I think movement_periods.reach_period is not present in the random-target sessions.

torch_brain/models/poyo.py
```diff
@@ -346,9 +329,9 @@ def __call__(self, data: Data) -> Dict:
             "latent_index": latent_index,
             "latent_timestamps": latent_timestamps,
             # output sequence
-            "output_session_index": pad8(session_index),
+            "output_session_index": pad8(output_session_index),
             "output_timestamps": pad8(output_timestamps),
-            "output_mask": pad8(output_mask),
+            "output_mask": track_mask8(output_session_index),
```
vinamarora8, line 334: I went with the `pad8` route because that way we can `&` the evaluation mask with it (if doing eval) and don't have to pass a separate eval mask to the validator.

vinamarora8:
I like the change overall. Left some comments wrt this code being able to run like before.

vinamarora8:
Also note we're still missing invalid_periods here. Leaving it up to you to decide whether to have a separate PR for that.

mazabou:
Sounds good, I wanted to make a quick draft of the overall concept. Since you agree with the change overall, I will go ahead and implement it! 

The `invalid_periods` is addressed through `cursor_outlier_segments: 0.0`.

vinamarora8:
> The `invalid_periods` is addressed through `cursor_outlier_segments: 0.0`. 

It was only outliers that had weight 0. Invalid trials had weight 0.1

mazabou:
This branch was tested and works as expected. Two things that need to be fixed are 1. bring back normalization to POYO from POYOPlus and 2. Report only on reach

PR #36, opened 2025-02-01
Update POYO/POYO+ in torch_brain

mazabou:
This PR applies the changes made to POYO model and scripts to POYOPlus


examples/poyo/configs/dataset/perich_miller_population_2018.yaml
```diff
@@ -81,8 +81,20 @@
...
+    readout:
+      readout_id: cursor_velocity_2d
+      normalize_mean: 0.
+      normalize_std: 20.
+      weights:
+        movement_phases.random_period: 1.0
+        movement_phases.hold_period: 0.1
+        movement_phases.reach_period: 5.0
+        movement_phases.return_period: 1.0
+        cursor_outlier_segments: 0.0
```
vinamarora8, line 93: Previous configurations of POYO had applied a weight of 0.1 to "invalid trials" (different from outliers). We're missing that here.

examples/poyo/train.py
```diff
@@ -295,7 +295,6 @@ def main(cfg: DictConfig):
         num_latents_per_step=cfg.model.num_latents,
         readout_spec=readout_spec,
         sequence_length=cfg.sequence_length,
-        subtask_weights=cfg.subtask_weights,
     )
```
vinamarora8, line 298: Not critical at all: I would prefer having a more transparent way to see (in code) how weights are being used in the training of the model. It has become very opaque now, with a torch_brain.utils function looking at them. Just pointing out. Not critical to address.

examples/poyo/train.py
```diff
@@ -338,8 +337,12 @@ def main(cfg: DictConfig):
         check_val_every_n_epoch=cfg.eval_epochs,
         max_epochs=cfg.epochs,
         log_every_n_steps=1,
+        strategy=(
+            "ddp_find_unused_parameters_true" if torch.cuda.is_available() else "auto"
+        ),
```
vinamarora8, line 342: Not critical again: I don't think we have any unused weights. So this option seems redundant. I have seen warnings that claim using this option unnecessarily can have an impact on iteration time.

vinamarora8 (commented):
Looks good overall. Only critical thing I found was the missing "invalid trials" weight. Left some minor comments that are not critical to address as this moment.

examples/poyo/configs/dataset/perich_miller_population_2018.yaml
mazabou: thanks for the catch. I will update this!

examples/poyo/train.py
mazabou: That's a fair point. I would argue that them being in the config file is sufficient? I would actually expect the average user to look at the config more than the actual train script

examples/poyo/train.py
mazabou: yep, you are correct this is only needed for POYO+ not POYO, since all modules are always used

examples/poyo/configs/dataset/perich_miller_population_2018.yaml
mazabou: I remember why I did not included it. It was actually never used in the perich_miller data. The invalid data was just considered "random"

Issue #39, opened 2025-02-07
Dataset.get() overwrites SessionDescription in Data objects


vinamarora8:
Looking at brainsets pipelines, the SessionDescription metadata is currently stored in the data.session field.
This line in `torch_brain/data/dataset.py` (the `get` function) overwrites this session metadata in the sliced sample Data object.

```
sample.session = recording_id
```
----

Additionally, we re-define the unit-id prefix at multiple places as `f"{data.brainset}/{data.session}/"`. 
Proposal: `recording_id` is already defined as `<brainset>/<session_id>`, we should use this at all places as the unit-id prefix, instead of re-working it in different functions.

mazabou:
This indeed an error. It should just be appending `brainset.id` to `data.session.id` like we are doing for the units. We would need to make sure to update this line in both POYO and POYOPlus: https://github.com/neuro-galaxy/torch_brain/blob/bea2c4f6e4c0316791565c309c769d8bd5773b99/torch_brain/models/poyo.py#L356

I've also found this other bug that I did not get around to fixing:
```
        sample.units.id = np.core.defchararray.add(
            f"{sample.brainset}/{sample.session}/", sample.units.id.astype(str)
        )
```

it should be:
```
        sample.units.id = np.core.defchararray.add(
            f"{sample.brainset.id}/{sample.session.id}/", sample.units.id.astype(str)
        )
```

---

Re: using recording_id which is always `<brainset>/<device_id>`, I still want to add the option of using either `<brainset>/<session_id>` or `<brainset>/<device_id>` (previously sortset_id). Open to figuring out a simple solution for all of this.

vinamarora8:
Fixed in #40

PR #40, opened 2025-02-09
Unify session/unit/subject ID prefixing logic

vinamarora8:
Addresses issues raised in #39 

1.  Unify id prefix applications into methods `_get_*_ids_with_prefix(data: Data) -> str` 
2. Add a way for users to control the prefixes by providing `unit_id_prefix_fn`, `session_id_prefix_fn`, and `subject_id_prefix_fn`

### DIFF
--- a/torch_brain/data/dataset.py
+++ b/torch_brain/data/dataset.py
@@ -60,6 +60,12 @@ class Dataset(torch.utils.data.Dataset):
             in a session based on a predefined split.
         transform: A transform to apply to the data. This transform should be a callable
             that takes a Data object and returns a Data object.
+        unit_id_prefix_fn: Optional[Callable[[Data], str]]
+            Function to generate prefix strings for unit IDs to ensure uniqueness across the dataset.
+            Takes a Data object as input and returns a string prefix.
+            By default (when None), uses: "<brainset-id>/<session-id>/"
+            Example:
+                >>> unit_id_prefix_fn = lambda data: f"{data.brainset.id}/{data.session.id}/"
     """
 
     _check_for_data_leakage_flag: bool = True
@@ -373,8 +379,11 @@ class Dataset(torch.utils.data.Dataset):
         return sorted(list(self.recording_dict.keys()))
 
     def _get_unit_ids_with_prefix(self, data: Data) -> np.ndarray:
-        r"""Add prefix string to data.units.id. This is an inplace modification.
-        Prefix string is chosen based on unit_id_prefix_fmt.
+        r"""Add prefix string to data.units.id and return a new numpy string array.
+
+        If `unit_id_prefix_fn` is set (not None), then this function is used
+        to decide the prefix string. Otherwise, the default value is:
+        `<data.brainset.id>/<data.session.id>/`
         """
         if self.unit_id_prefix_fn is not None:
             prefix_str = self.unit_id_prefix_fn(data)
--- a/torch_brain/data/dataset.py
+++ b/torch_brain/data/dataset.py
@@ -25,6 +25,11 @@ class DatasetIndex:
     end: float
 
 
+default_session_id_prefix_fn = lambda data: f"{data.brainset.id}/"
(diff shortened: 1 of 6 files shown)


torch_brain/data/dataset.py
```diff
@@ -371,52 +374,31 @@ def get_recording_config_dict(self):
...
+        r"""Add prefix string to data.units.id and return a new numpy string array.
+
+        If `unit_id_prefix_fn` is set (not None), then this function is used
+        to decide the prefix string. Otherwise, the default value is:
+        `<data.brainset.id>/<data.session.id>/`
+        """
+        if self.unit_id_prefix_fn is not None:
+            prefix_str = self.unit_id_prefix_fn(data)
+        else:
+            prefix_str = f"{data.brainset.id}/{data.session.id}/"
```
mazabou, line 391: can we make this a `default_unit_id_prefix_fn`?

torch_brain/data/dataset.py
mazabou, line 1: what about the session_id, are we imagining something similar to `unit_id_prefix_fn`? because instead of using `brainset_id/session_id` it could be `brainset_id/subject_id` or `brainset_id/device_id`

mazabou (commented):
I think we need to think a bit more about variable names. It's currently a bit confusing. I left a few comments, would love to see what you think

torch_brain/models/poyo.py
```diff
@@ -375,7 +375,7 @@ def __call__(self, data: Data) -> Dict:
         }
 
         if self.eval:
-            batch["session_id"] = data.session
+            batch["session_id"] = data.recording_id
```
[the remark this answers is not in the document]
vinamarora8: Okay so here is my thought process: The `Dataset` object only cares about recordings (or `Data` h5 files). It does not care to make any assumption about the relationship between these recordings and sessions. It could totally be possible that a session is split across multiple recordings.

However, in POYO and POYO+, _we make this assumption/assertion_ that all sessions are kept in separate h5 files. And so, these implementations treat `recording_id` as `session_id`.

---

It could be good for us to consider alternate names for `recording_id`. Maybe `data_file_id` is more accurate? (Doesn't need to introduce a new concept - "recordings")

torch_brain/data/dataset.py
vinamarora8: Yeah sure, that'll be cleaner. Where would that be defined? A `torch_brain/data/dataset.py` module level variable perhaps?

torch_brain/data/dataset.py
vinamarora8: You raise a fair point. Every project might need a different definition of "session_id".  I think a natural way for this is the following: 

We remove `recording_id` completely, and compute `session_id` manually **inside the tokenizer**:

```python
class POYOTokenizer:
    ...
    def __call__(self, data: Data) -> dict:
        ....
        session_id  = f"{data.brainset.id}/{data.session.id}"
        output_session_index = self.session_tokenizer(session_id)
        ...
        if self.eval:
            batch["session_id"] = session_id     
````

vinamarora8:
> I think we need to think a bit more about variable names. It's currently a bit confusing. I left a few comments, would love to see what you think

Very fair points. Here are my suggestions overall:
1. We remove `recording_id` and compute `session_id` inside the tokenizer.
2. We rename `recording_id` to something more easily understandable: `file_id` or `data_id` (more general, once we add the ability to create `Dataset` from `Data` objects)

mazabou:
> Hey @vinamarora8 @mazabou I am currently working on a PR to have access to `sample.session.id` with the following way
> 
> `sample.session.id = f"{data.brainset.id}/{data.session.id}"`
> 
> Should I wait for the end of your PR before opening one on the subject?

I would suggest making the edit in a different PR. It seems like this PR might make major changes in the API, and might take a bit longer to merge.

mazabou (approved):
Ok I think this is okay for now. Let's merge

PR #41, opened 2025-02-10
RFC: A new way to initialize `Dataset`

vinamarora8:
This is a request for comments:

I tried adding a "Initialize Dataset from a list of Data objects," and realized our current implementation of `Dataset.__init__()` is getting quite hairy. So came up with this alternate way of initializations: through `@classmethods` only. Users will have access to 3 (or more in the future) ways of initializing datasets:

1. `Dataset.from_config(root, config)`
2. `Dataset.from_data_files(filepath_dict)`
3. `Dataset.from_data_objects(data_dict)`

One issue I see is that the lack of `Dataset.__init__` could be worrying for more novice users. For this, we can choose one of the three ways to act as the `__init__` way (I think `from_data_files` is the most novice-user friendly?).

### DIFF
--- a/torch_brain/data/dataset.py
+++ b/torch_brain/data/dataset.py
@@ -1,3 +1,4 @@
+from __future__ import annotations
 import copy
 import logging
 import os
@@ -14,6 +15,7 @@ import numpy as np
 import torch
 
 from temporaldata import Data, Interval
+from torch_brain.transforms import TransformType
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


torch_brain/data/dataset.py
```diff
@@ -68,51 +70,95 @@ class Dataset(torch.utils.data.Dataset):
...
     """
 
+    split: Optional[str] = None
+    transform: Optional[TransformType] = None
+    unit_id_prefix_fn: Optional[Callable[[Data], str]] = (None,)
     _check_for_data_leakage_flag: bool = True
     _open_files: Optional[Dict[str, h5py.File]] = None
     _data_objects: Optional[Dict[str, Data]] = None
 
-    def __init__(
```
ShivashriganeshMahato, line 75: Restructuring LGTM. I agree that we should keep one of the factories implemented for `__init__`, and `from_data_files` is most straightforward.

mazabou:
Like we discussed yesterday, I think introducing all these different ways of defining a dataset, when not all of them will have the same methods makes this not fit to be all in the same class. Currently, Dataset assumes that there are brainsets, and subject descriptions and units etc.. 

We can define a "simple" dataset interface for cases where we want to use `Dataset.from_data_objects(data_dict)` this should be a few lines of code with the only function needed to be defined being get. The same samplers we have in torch_brain will work.

mazabou:
Closing this draft. I do not thing we need to add multiple ways of defining Dataset object unless we're able to support the same functionality across the board.

PR #43, opened 2025-02-10
Fix access to subject or unit

AlexandreAndr:
Want to be able to access the subject id according to the convention: `brainset.id/subject.id`

As done in this function
```
    def get_subject_ids(self):
        r"""Returns all subject ids in the dataset."""
        subject_ids = []
        for recording_id in self.recording_dict.keys():
            data = self._data_objects[recording_id]
            subject_ids.append(f"{data.brainset.id}/{data.subject.id}")
        return sorted(list(set(subject_ids)))
```

### DIFF
--- a/torch_brain/data/dataset.py
+++ b/torch_brain/data/dataset.py
@@ -284,8 +284,9 @@ class Dataset(torch.utils.data.Dataset):
         sample = data.slice(start, end)
 
         sample.units.id = np.core.defchararray.add(
-            f"{sample.brainset}/{sample.session}/", sample.units.id.astype(str)
+            f"{sample.brainset.id}/{sample.session.id}/", sample.units.id.astype(str)
         )
+        sample.subject.id = f"{data.brainset.id}/{data.subject.id}"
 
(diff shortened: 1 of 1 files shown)


AlexandreAndr:
The question is should `get_recording_data` be also updated

torch_brain/data/dataset.py
```diff
@@ -286,6 +286,7 @@ def get(self, recording_id: str, start: float, end: float):
         sample.units.id = np.core.defchararray.add(
             f"{sample.brainset}/{sample.session}/", sample.units.id.astype(str)
```
mazabou, line 287: Since you are at it, I would also fix this error:
```suggestion
            f"{sample.brainset.id}/{sample.session.id}/", sample.units.id.astype(str)
```

mazabou (approved):
lgtm, I suggested a bug fix

mazabou:
> The question is should `get_recording_data` be also updated

If we are doing it for unit_ids, then we should do it for subject_id

PR #45, opened 2025-02-11
Merging tokenizer with model

mazabou:
The idea of attaching a tokenizer to the model is already used in our code (InfiniteVocabEmbedding), and given that we keep passing object references across the model and the tokenizer, and that each tokenizer is specific to a given model, we can merge the tokenizer and make it a method of the model. This also has the added advantage of simplifying the loading of the model from a checkpoint. 

This draft PR tests what this might look like. 

- We need to stick with either tokenize or tokenizer (and update InfiniteVocabEmbedding).
- We shouldn't use "batch" inside of the tokenizer, it's actually a single sample
- The tokenizer is defined to return exactly the input to the forward function of the model, maybe we can have it return the outputs and then any extras used outside of the model. this is a tentative idea, but we should try to simplify/unify the interface.
- A question is whether passing this tokenizer to the cpu workers will cause issues with pickling or making copies of the models which is on GPU. this needs to be tested.

### DIFF
--- a/examples/poyo/train.py
+++ b/examples/poyo/train.py
@@ -17,7 +17,7 @@ from omegaconf import DictConfig, OmegaConf
 from temporaldata import Data
 
 from torch_brain.registry import MODALITIY_REGISTRY, ModalitySpec
-from torch_brain.models.poyo import POYOTokenizer, poyo_mp
+from torch_brain.models.poyo import poyo_mp
 from torch_brain.utils import callbacks as tbrain_callbacks
 from torch_brain.utils import seed_everything
 from torch_brain.utils.stitcher import DecodingStitchEvaluator
@@ -77,6 +77,7 @@ class TrainWrapper(L.LightningModule):
         }
 
     def training_step(self, batch, batch_idx):
+        batch, extra = batch
         target_values = batch.pop("target_values")
         target_weights = batch.pop("target_weights")
         output_mask = batch.pop("output_mask")
@@ -116,11 +117,12 @@ class TrainWrapper(L.LightningModule):
         return loss
 
     def validation_step(self, batch, batch_idx):
+        batch, extra = batch
         target_values = batch.pop("target_values")
         batch.pop("target_weights")
-        absolute_starts = batch.pop("absolute_start")
-        session_ids = batch.pop("session_id")
-        output_mask = batch.pop("output_mask")
+        absolute_starts = extra.pop("absolute_start")
+        session_ids = extra.pop("session_id")
+        eval_mask = extra.pop("eval_mask")
 
         # forward pass
         output_values = self.model(**batch)
@@ -129,7 +131,7 @@ class TrainWrapper(L.LightningModule):
         batch["target_values"] = target_values
         batch["absolute_start"] = absolute_starts
         batch["session_id"] = session_ids
-        batch["output_mask"] = output_mask
+        batch["eval_mask"] = eval_mask
 
         return output_values
 
@@ -158,13 +160,11 @@ class DataModule(L.LightningDataModule):
         )
         self.train_dataset.disable_data_leakage_check()
 
-        eval_tokenizer = copy.copy(self.tokenizer)
-        eval_tokenizer.eval = True
         self.val_dataset = Dataset(
             root=self.cfg.data_root,
             config=self.cfg.dataset,
             split="valid",
-            transform=eval_tokenizer,
+            transform=self.tokenizer,
         )
         self.val_dataset.disable_data_leakage_check()
 
@@ -172,7 +172,7 @@ class DataModule(L.LightningDataModule):
             root=self.cfg.data_root,
             config=self.cfg.dataset,
             split="test",
-            transform=eval_tokenizer,
+            transform=self.tokenizer,
         )
         self.test_dataset.disable_data_leakage_check()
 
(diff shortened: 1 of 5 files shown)


torch_brain/models/poyo.py
```diff
@@ -73,12 +77,37 @@ def __init__(
         lin_dropout=0.4,
         atn_dropout=0.0,
         emb_init_scale=0.02,
+        sequence_length: float = 1.0,
```
milosobral, line 80: There seems to be some disconnect between the docstring and the actual attributes

torch_brain/models/poyo.py
```diff
@@ -132,10 +161,87 @@ def __init__(
...
+        )
+        spike_unit_index = np.concatenate([se_unit_index, spike_unit_index])
+        spike_timestamps = np.concatenate([se_timestamps, spike_timestamps])
+
+        # unit_index is relative to the recording, so we want it to map it to
+        # the global unit index
+        local_to_global_map = np.array(self.unit_emb.tokenizer(unit_ids))
+        spike_unit_index = local_to_global_map[spike_unit_index]
+
+        ### prepare latents
```
milosobral, line 196: Maybe we can think about having a POYOModel Base class which also implements these portions of code that will likely be reused in every model's tokenizer. Might also make it easier to write tests that apply to all models and to make sure training code stays very consistent.

torch_brain/models/poyo.py
```diff
@@ -132,10 +161,87 @@ def __init__(
...
+
+        output_timestamps, output_values, output_weights, eval_mask = (
+            prepare_for_readout(data, self.readout_spec)
+        )
+
+        # create session index for output
+        output_session_index = self.session_emb.tokenizer(data.session)
+        output_session_index = np.repeat(output_session_index, len(output_timestamps))
+
+        batch = {
```
milosobral, line 212: You mentionned it in the comment on this PR but I agree that calling this batch is definitely confusing

milosobral (commented):
Really like the idea of formally coupling Tokenizer and Model as the input of one specifically has to match the ouput of the other. 
As I mentionned in a comment I think it might therefore be a good idea to enforce the structure and presence of that tokenizer with a base class.

torch_brain/models/poyo.py
```diff
@@ -132,10 +161,87 @@ def __init__(
...
+                if eval_mask is not None
+                else track_mask8(output_session_index)
+            ),
+        }
+
+        return batch, extra
+
+    def get_tokenizer(self):
+        return self.tokenize
+
```
vinamarora8, line 244: I don't see a point of this function yet. What's your philosophy behind not using `model.tokenize` directly as the tokenizer transform?

examples/poyo/train.py
```diff
@@ -116,11 +117,12 @@ def training_step(self, batch, batch_idx):
...
     def validation_step(self, batch, batch_idx):
+        batch, extra = batch
         target_values = batch.pop("target_values")
         batch.pop("target_weights")
-        absolute_starts = batch.pop("absolute_start")
-        session_ids = batch.pop("session_id")
-        output_mask = batch.pop("output_mask")
+        absolute_starts = extra.pop("absolute_start")
+        session_ids = extra.pop("session_id")
+        eval_mask = extra.pop("eval_mask")
```
vinamarora8, line 125: While we are trying to improve the model interfaces, I have a better solution to avoid this "pop from dictionary and put it back" thing that we do inside train/validation steps. 

Recently, I've been organizing my batch dictionaries as:
```python
batch = {
    "model_input": { ... }, # everything needed for the model forward pass
    "extra_stuff_1": ...,
    "extra_stuff_2": ...,
}
```

Then, inside train/val step, we can simply do:
```python
model(**batch["model_input"])
```
And the "extra_stuff_*" can stay in the `batch` dictionary.

If this text doesn't make too much sense, I would be happy to implement it as a PR.

vinamarora8 (changes requested):
I'm glad we're finally considering merging model and tokenizer into the same class.
I'm looking at this PR mostly as introducing a stronger API for torch_brain models, and I think it's mostly the correct approach. Left 2 comments. For me, both comments are critical to be resolved before merging.

torch_brain/models/poyo.py
vinamarora8: Totally agree with having a well defined base class. I think all it needs to be is this?

```python
class TorchBrainModel(nn.Module):

    def tokenizer(self, data: Data) -> Any:
        raise NotImplementedError
        
    def forward(self, *args, **kwargs):
        raise NotImplementedError
```

examples/poyo/train.py
mazabou: agreed with this approach. Although `model_input` includes output_tokens for example which might be a bit confusing? Could we come up with another more general term as a convention?

torch_brain/models/poyo.py
mazabou: will need to update

torch_brain/models/poyo.py
mazabou: Most variants of POYO will indeed share a lot of the same code, but each model (MLP, TCN, NDT, MtM...) will likely have its own tokenizer.

examples/poyo/train.py
vinamarora8: `output_tokens` being part of `model_input` is a side-effect of us naming POYO-forward's arguments as output_tokens. Maybe we should change the names of forward args to be `query_tokens` or smth...

mazabou: also, some models might want to compute their loss inside of the forward, so things like target or output could be included

vinamarora8: Re: your second comment. In that case output is _really_ a part the model's input, so it's fine. I don't see any confusion.

mazabou:
Closing this. This is now implemented in #51

PR #47, opened 2025-02-11
Add unit filering (based on keywords) in transforms

AlexandreAndr:


torch_brain/transforms/unit_filter.py
```diff
@@ -0,0 +1,56 @@
...
+            target_obj.unit_index = relabel_map[target_obj.unit_index]
+        elif isinstance(target_obj, RegularTimeSeries):
+            assert len(nested_attr) == 2
+            setattr(
+                target_obj,
+                nested_attr[1],
+                getattr(target_obj, nested_attr[1])[:, keep_unit_mask],
+            )
+        else:
+            raise ValueError(f"Unsupported type for {self.field}: {type(target_obj)}")
```
vinamarora8, line 54: We're also using this exact masking code in UnitDropout. I think we should refactor it out into a maybe a utils function (`apply_units_mask(data, mask)` or smth)

torch_brain/transforms/unit_filter.py
```diff
@@ -0,0 +1,73 @@
...
+    def filter_fn(units: ArrayDict) -> np.ndarray:
+        unit_ids = units.id
+        no_keyword_unit_mask = np.char.find(unit_ids, keyword) != -1
+
+        if drop == False:
+            return ~no_keyword_unit_mask
+
+        return no_keyword_unit_mask
+
+    return filter_fn
```
vinamarora8, line 73: This should be implemented as a class.

```python
class KeywordBasedUnitMask:
    r"""Generate a unit mask based on the presence/absence of a keyword in the unit id.
    """
    
    def __init__(self, keyword: str, drop: bool = False):
        r"""
        Args:
            keyword: the keyword to search for in the unit IDs
            drop: If False, units are dropped when keyword is _present_ in unit id.
                If True, units are dropped when keyword is _absent_ in unit id.
        """
        self.keyword = keyword
        self.drop = drop
        
    def __call__(self, units: ArrayDict) -> np.ndarray:
        unit_mask = np.char.find(units.id, keyword) != -1

        if self.drop:
            unit_mask = ~unit_mask

        return unit_mask
```

torch_brain/transforms/unit_filter.py
```diff
@@ -0,0 +1,73 @@
+from typing import Callable
+
+import numpy as np
+
+from temporaldata import ArrayDict, Data, IrregularTimeSeries, RegularTimeSeries
+
+
+class FilterUnit:
```
vinamarora8, line 8: This is a class, and so should not be named with an action-phrase. `UnitFilter` seems better. (sry for being nit picky)

torch_brain/transforms/unit_filter.py
```diff
@@ -0,0 +1,73 @@
...
+    r"""
+    Drop units based on the `filter_fn` given in the constructor.
+    This operation is performed in-place.
+    """
+
+    def __init__(
+        self,
+        filter_fn: Callable[[np.ndarray], np.ndarray],
```
vinamarora8, line 16: `filter_fn` -> `mask_fn` because the function gives you a mask, and doesn't do the filtering itself.

torch_brain/transforms/unit_filter.py
```diff
@@ -0,0 +1,73 @@
...
+        self.field = field
+        self.filter_fn = filter_fn
+        self.reset_index = reset_index
+
+    def __call__(self, data: Data) -> Data:
+        # get units from data
+        unit_ids = data.units.id
+        num_units = len(unit_ids)
+
+        unit_mask = ~self.filter_fn(data.units)
```
vinamarora8, line 35: I think we should follow the convention "mask = True means the unit is kept". So here you should not flip the mask returned by `filter_fn(...)`.  (cc @mazabou)

vinamarora8 (changes requested):
Thanks for making the edit, and adding a test! I like the direction, but it can be made better. See my comments below.

torch_brain/transforms/unit_filter.py
vinamarora8: Also the input to the fn is a `ArrayDict`, so the type definition needs to be fixed.

torch_brain/transforms/unit_filter.py
AlexandreAndr: Agree it is not elegant to have to use a not

vinamarora8 (dismissed):

vinamarora8 (approved):
Changes look good. I've made small edits, mainly to comments and to the test. Thanks for adding a feature and a test! Merging.

vinamarora8:
@mazabou github won't let me merge because I pushed the last commit. Can you merge instead?

AlexandreAndr:
let me try to squash commit to see if this work :)

AlexandreAndr:
we definitely need Mehdi 😅

vinamarora8 (dismissed):

torch_brain/transforms/unit_filter.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class KeywordBasedUnitMask:
+    r"""Generate a unit mask based on the presence/absence of a keyword in the unit id.
+
+    Args:
+        keyword: the keyword to search for in the unit ids
+        drop: If False, units are dropped when keyword is _absent_ in unit id.
+            If True, units are dropped when keyword is _present_ in unit id.
+    """
```
mazabou, line 64: This is a hard one. Not sure if this should be a function or a class.

vinamarora8: We want to instantiate it from the config, so it has to be a class.

AlexandreAndr: About the regex, I see the interest even if from my experience not everyone is familiar with it. Do you think, it would be relevant to have another class allowing to use regex or should everything be supported on this one?

mazabou: Ok, then I would rather this class inherit from `UnitFilter` and have an extra field for the keyword. Sorry @AlexandreAndr I realize that that's what you originally had.

mazabou: Actually the regex pattern for the keyword is the keyword itself. So it will be pretty easy even for people who don't know how to regex works

torch_brain/transforms/unit_filter.py
```diff
@@ -0,0 +1,76 @@
...
+        self.keyword = keyword
+        self.drop = drop
+
+    def __call__(self, units: ArrayDict) -> np.ndarray:
+        unit_mask = np.char.find(units.id, self.keyword) != -1
+
+        if self.drop:
+            unit_mask = ~unit_mask
+
+        return unit_mask
```
mazabou, line 76:
```suggestion
import re
from typing import Pattern, Union

class UnitFilterById(UnitFilter):
    r"""...
    
    Args:
        pattern: The regex pattern to match against the unit ids.
        field: str,
        reset_index: bool = True,
    """

    def __init__(self, pattern: Union[str, Pattern], field, reset_index):
        self.pattern = re.compile(pattern) if isinstance(pattern, str) else pattern
        super().__init__(self._generate_unit_mask, field, reset_index)

    def _generate_unit_mask(self, units: ArrayDict) -> np.ndarray:
        unit_mask = np.array([bool(self.pattern.search(uid)) for uid in units.id])
        return unit_mask
```

torch_brain/transforms/unit_filter.py
```diff
@@ -0,0 +1,76 @@
...
+from temporaldata import ArrayDict, Data, IrregularTimeSeries, RegularTimeSeries
+
+
+class UnitFilter:
+    r"""
+    Drop units based on the `mask_fn` given in the constructor.
+    This operation is performed in-place.
+
+    Args:
+        mask_fn: A function that takes the unit ids and returns a mask to keep the units.
```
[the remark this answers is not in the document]
AlexandreAndr: We could but I don't know if we should implement this on this PR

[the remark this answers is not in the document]
AlexandreAndr: What is your thoughts? Is it a top priority for you?

torch_brain/transforms/unit_filter.py
```diff
@@ -0,0 +1,83 @@
...
+        reset_index (bool, optional): If True, it will reset_index the unit index of the time series.
+        drop (bool, optional): If True, units matching the pattern will be dropped.
+    """
+
+    def __init__(
+        self,
+        pattern: Union[str, Pattern],
+        field: str,
+        reset_index: bool = True,
+        drop: bool = True,
```
mazabou, line 73: I am conflicted by this argument. it's not easy to understand. we can call it something more explicit like `keep_matches`. Also, by default i would think drop would be False. if i am "filtering by id" i am expecting that i will keep the ids i am matching by default.

mazabou (changes requested):
Small edit suggestion. 

Can you also git merge with the latest main?  the testing is failing because of brainsets failing to download the data for some reason.

mazabou (approved):
LGTM! Thanks Alex

torch_brain/transforms/unit_filter.py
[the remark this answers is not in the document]
mazabou: nope, we can do this in a different PR in the future

PR #48, opened 2025-02-11
refractor NDT-2 

mazabou:
The idea is to make NDT-2 more readable and share the same ideas used for other models in torch_brain.
This is just a first attempt at simplifying things. The main idea is that no tensor should be built on GPU, everything should be done on CPU in the tokenizer including masking. Additionally, we want to move away from the current decoder which is currently specific to a task, to use the prepare_for_readout function in torch_brain.utils. 

This is just a draft of where we want things to go.


mazabou:
Closing this. New PR https://github.com/neuro-galaxy/torch_brain/pull/76

PR #49, opened 2025-02-11
replace Lamb with SparseLamb

mazabou:
This PR replaces the Lamb optimizer, with SparseLamb in both POYO and POYO+. This applies to any model that won't use all parameters in every single train step. Typically this will happen with InfiniteVocabEmbedding and MultiTaskReadout.


vinamarora8:
Hey, I think there's a small bug hiding inside our SparseLamb implementation. It assumes gradient = 0 means that embedding wasn't used and so it would skip updating its momentum etc. However, there's a tiny chance that an embedding was used, and its gradient does came out to be zero naturally. In that case, we should be updating its momentum etc. even if grad is 0.

I have another implementation of SparseLamb in one of my projects that takes care of this edge case. Would be happy to share if you're interested in fixing this problem.

AlexandreAndr (dismissed):

mazabou:
> Hey, I think there's a small bug hiding inside our SparseLamb implementation. It assumes gradient = 0 means that embedding wasn't used and so it would skip updating its momentum etc. However, there's a chance that an embedding was used, and its gradient does came out to be zero naturally. In that case, we should be updating its momentum etc. even if grad is 0.
> 
> I have another implementation of SparseLamb in one of my projects that takes care of this edge case. Would be happy to share if you're interested in fixing this problem.

Indeed, I am making the assumption that the detection is done through grad=0. The best compromise I found was to isolate the targeted modules, and only do this for those modules to minimize false positives. If you have an alternative implementation, please do share it!

mazabou:
Another test worth doing is the likelihood of having exactly 0, isn't that almost impossible?

vinamarora8:
Here is my implementation of SparseAdamW in another project. I believe this is "correct" in the sense that it makes no assumption about grad=0 implying no access. 

- `SparseInfiniteVocabEmbedding`[(link)](https://github.com/nerdslab/nujepa/blob/main/src/models/embedding.py): You need to replace `InfiniteVocabEmbedding` with `SparseInfiniteVocabEmbedding` if you want to use `SparseAdamW`. What the sparse version does is: in every forward call, it notes down which embeddings were accessed, and stores is as a property of the parameter. Distributed case has been taken care of in this implementation.

- `SparseAdamW` [(link)](https://github.com/nerdslab/nujepa/blob/main/src/sparseadamw.py): Only updates the embedding rows that were accessed. In this link, modifications to AdamW start at line 308.

mazabou:
This seems to only apply to AdamW though, we would need to make an equivalent implementation of SparseLamb.

vinamarora8:
> This seems to only apply to AdamW though, we would need to make an equivalent implementation of SparseLamb.

Yes, I think it should be ~trivial to apply the same principle to Lamb. If you want it in this project, I would be happy to do so.

mazabou:
two questions:
- why not use `sparse=True` in `F.embedding`?
- why not use [SparseAdam](https://pytorch.org/docs/stable/generated/torch.optim.SparseAdam.html)?

vinamarora8:
Yeah these torch builtins are more appealing than something custom.
So you're suggesting we just use SparseAdam for POYO/POYO+ embedding tables? We could try it out.
It might be fine too; there's a meta of not having weight decay on nn.Embedding https://stats.stackexchange.com/questions/576463/why-not-perform-weight-decay-on-layernorm-embedding

vinamarora8:
Since this PR changes POYO’s training dynamics by switching from Lamb to SparseLamb, it would be good to keep a record of the improvements. Especially for POYO, given the paper uses normal Lamb. 

Could you upload CSV files or screenshots of test results comparing both optimizers? We can keep it as a record in this PR's comments section. It doesn't need to be exhaustive; POYO-MP test results on the MP dataset would suffice. The goal is just to document changes for future reference, and I think it’s a good practice for such PRs.

PR #52, opened 2025-02-13
Make `DecodingStitchEvaluator` a standalone Class.

vinamarora8:
**The Problem**: Existing implementation of `DecodingStitchEvaluator` is a `L.Callback`. This has 2 problems: 1) it can only be used when the project uses Lightning, and 2) Passing non-trivial data (DataForDecodingStitchEvaluator) to a callback obscures the program and data flow.
We don't want to restrict users to Lightning unless absolutely necessary, and we should strive to make the program/data flow as visible as possible while ensuring simplicity.

**This PR**: Converts `DecodingStitchEvaluator` to a standalone class. With the following API:
```python
# Initialization. This sets up the metric functions, and the cache. Same as current impl.
stitch_evaluator = DecodingStitchEvaluator(session_ids=..., modality_spec=...)

# update: This updates the cache store. 
# Would be called at the end of each validation/test batch
stitch_evaluator.update(
    timestamps=..., # FloatTensor, [B, N]
    preds=..., # Tensor, [B, N, D]
    targets=..., # Tensor, [B, N, D]
    eval_masks=..., # BoolTensor, [B, N]
    session_ids=..., # List[str], with length = B
    absolute_starts=..., # FloatTensor, [B,]
)

# compute: Stitch, compute metrics, and return metric dictionary.
# Would be called at the end of validation/test epoch
metric_dict = stitch_evaluator.compute()
stitch_evaluator.reset()  # <- reset: Resets the cache, so we're ready for next validation epoch
```

This API is very similar to the one followed by `torchmetrics`.

Another design decision: I have moved metric logging logic into the train script itself. I couldn't find a clean way to hide this detail into a utils fn, since it accesses `self.trainer`, and the `TrainWrapper` itself for logging functions.

### TODO:
- [x] Mirror changes to POYO+
- [x] Add tests for `DecodingStitchEvaluator`


mazabou:
This is great!

examples/poyo/train.py
```diff
@@ -123,10 +138,31 @@ def validation_step(self, batch, batch_idx):
...
+
+        if self.trainer.is_global_zero:
+            logging.info(f"Logged {len(metric_dict_with_prefix)} {prefix} metrics.")
+            rprint(metric_df)
+
+            for logger in self.trainer.loggers:
+                if isinstance(logger, L.pytorch.loggers.TensorBoardLogger):
+                    logger.experiment.add_text(
+                        f"{prefix}_metrics", metric_df.to_markdown()
+                    )
```
mazabou, line 161: in this particular script, we do not support tensorboard, so should we remove this?

torch_brain/utils/stitcher.py
```diff
@@ -148,265 +171,200 @@ def __init__(
...
-
-
-class MultiTaskDecodingStitchEvaluator(L.Callback):
-    def __init__(self, metrics: dict):
+class MultiTaskDecodingStitchEvaluator:
+    def __init__(
+        self,
+        metrics: dict,
+        sequence_index: torch.Tensor,
+        device: torch.device,
```
mazabou, line 258: is there a reason for having a device arg here, but not in `DecodingStitchEvaluator`

torch_brain/utils/stitcher.py
```diff
@@ -148,265 +171,200 @@ def __init__(
...
         self,
-        trainer: L.Trainer,
-        pl_module: L.LightningModule,
-        data: DataForMultiTaskDecodingStitchEvaluator,
-        *args,
-        **kwargs,
+        timestamps: torch.Tensor,
+        preds: List[Dict[str, torch.Tensor]],
+        targets: List[Dict[str, torch.Tensor]],
+        decoder_indices: torch.LongTensor,
```
mazabou, line 270: for consistency, do we want to call this readout_indices

examples/poyo/train.py
```diff
@@ -108,13 +108,28 @@ def training_step(self, batch, batch_idx):
...
+
+    def test_step(self, batch, batch_idx):
+        self._eval_step(batch)
+
+    def on_test_epoch_end(self):
+        metric_dict = self.stitch_evaluator.compute()
+        self.stitch_evaluator.reset()
+        self._log_metric_dict(metric_dict, prefix="test")
+
+    def _eval_step(self, batch: Dict):
```
mazabou, line 126: let's call it `_evaluation_step` ? since it's `validation` and not `val` or `valid`.

examples/poyo_plus/train.py
```diff
@@ -131,13 +124,27 @@ def training_step(self, batch, batch_idx):
         return loss
 
     def validation_step(self, batch, batch_idx):
+        self._eval_step(batch, self.trainer.datamodule.val_evaluator)
```
mazabou, line 127: same here. `_evaluation_step`

examples/poyo_plus/train.py
mazabou, line 1: not a big fan of the major conceptual differences between POYO and POYO+

mazabou (commented):
This looks great overall! 

I have issues conceptually when it comes to difference between the two stitchers. 
I need a bit more time to think through this again.

examples/poyo/train.py
vinamarora8: yes, we should

torch_brain/utils/stitcher.py
vinamarora8: Cache is built on CPU, but torchmetrics' compute methods need to run on GPU. So in the `MultiTaskDecodingStitchEvaluator.compute` method, we do a `<metric>.to(device)` before computing the metric. We need to know what GPU to put the metric on, and that is why the constructor asks for specifying the `device`.

In earlier code, device is taken from the TrainWrapper Lightning Module. Because it was a callback, it had access to this information.

torch_brain/utils/stitcher.py
vinamarora8: yes, I will do that

examples/poyo_plus/train.py
vinamarora8: I can make the two train scripts look similar, i.e. make both look like POYO+'s train script.

torch_brain/utils/stitcher.py
vinamarora8: I am not a fan of building the cache on CPU by default (should be configurable somehow). But this is what the implementation was before.

Without changing this functionality, I can hide `device` being in the constructor by using the device of the first tensor received.

vinamarora8: Actually, it is much cleaner to hide this detail. Implemented in new commits.

mazabou:
Ok, hear me out. This PR made me realize how to make this cleaner and more general. You implemented this to mimic the Metric class in torchmetrics, which made me realize that this is basically a "MetricGroup", we can move in this direction even more.

Say you have a MetricGroup class that combines multiple metrics (let's say across multiple sessions), and calls them together through the same api. Here's how I would define it:

```
>>> metric_group = MetricGroup({"recording_1": torchmetrics.R2Score(), "recording_2": torchmetrics.R2Score(),})
>>> preds = {"recording_1": torch.rand(10, 1), "recording_2": torch.rand(20, 1)}
>>> targets = {"recording_1": torch.rand(10, 1), "recording_2": torch.rand(20, 1)}
>>> metric_group.update(preds, targets)
>>> metric_group.compute()
```

This makes it even closer to the OG torchmetrics class, just with dicts. Note that the main design choice here it to try to make the metrics dict mimic the `preds` and `targets` dict, so it's easier to understand.

If you have multiple tasks, you just add another depth level to your metrics and in the same way to your `preds` and `targets`!

```
>>> metrics = {
...     "recording_1": {
...         "task_1": torchmetrics.Accuracy(),
...         "task_2": torchmetrics.F1Score()
...     },
...     "recording_2": {
...         "task_1": torchmetrics.Accuracy(),
...     }
... }
>>> metric_group = MetricGroup(metrics)
>>> preds = {
...     "recording_1": {
...         "task_1": torch.rand(10, 1),
...         "task_2": torch.rand(20, 1)
...     },
...     "recording_2": {
...         "task_1": torch.rand(10, 1),
...     }
... }
>>> targets = {
...     "recording_1": {
...         "task_1": torch.rand(10),
...         "task_2": torch.rand(20)
...     },
...     "recording_2": {
...         "task_1": torch.rand(10),
...     }
... }
>>> metric_group.update(preds, targets)
>>> metric_group.compute()
```

Finally, you can have a third layer, for using multiple metrics on the same data -- here the symmetry between the metrics dict and targets/preds break because metrics has depth 3 but targets/preds has depth 2...

```
>>> metrics = {
...     "recording_1": {
...         "task_1": {
...             "metric_1": torchmetrics.Accuracy(),
...             "metric_2": torchmetrics.F1Score()
...         },
...         "task_2": {
...             "metric_1": torchmetrics.F1Score()
...         }
...     },
...     "recording_2": {
...         "task_1": {
...             "metric_1": torchmetrics.Accuracy(),
...         }
...     }
... }
>>> metric_group = MetricGroup(metrics)
>>> preds = {
...     "recording_1": {
...         "task_1": torch.rand(10, 1),
...         "task_2": torch.rand(20, 1)
...     },
...     "recording_2": {
...         "task_1": torch.rand(10, 1),
...     }
... }
>>> targets = {
...     "recording_1": {
...         "task_1": torch.rand(10),
...         "task_2": torch.rand(20),
...     },
...     "recording_2": {
...         "task_1": torch.rand(10),
...     }
... }
>>> metric_group.update(preds, targets)
>>> metric_group.compute()
```

Basically, you would have three possible depths: `recording_id > task_name > unit_name`. All the extra stuff like masking or metric building can be done outside of this `MetricGroup` and is basically model specific. This API is so easy to understand imo, and can be used with any objective (not just decoding tasks!)

--- 

Ok, the next concept is stitching. This just requires adding an extra argument to update which is `timestamps` (adding absolute_start should happen outside of this function!)

This is what it can look like
```
>>> metrics = {
...     "recording_1": torchmetrics.Accuracy(),
...     "recording_2": torchmetrics.Accuracy(),
... }
>>> metric_group = MetricGroup(metrics, enable_stitching=True)
>>> preds = {"recording_1": torch.rand(10, 1), "recording_2": torch.rand(20, 1)}
>>> targets = {"recording_1": torch.rand(10, 1), "recording_2": torch.rand(20, 1)}
>>> timestamps = {"recording_1": torch.rand(10, 1), "recording_2": torch.rand(20, 1)}
>>> metric_group.update(preds, targets, timestamps)
>>> metric_group.compute()
```

I think this is still very readable / intuitive, and not model specific. 

By default, this will cache everything until the end until you finish the evaluation epoch. If you want to leverage the cache flushing mechanism, it's simple!

```
>>> metrics = {
...     "recording_1": torchmetrics.Accuracy(),
...     "recording_2": torchmetrics.Accuracy(),
... }
>>> metric_group = MetricGroup(metrics, enable_stitching=True)
>>> metric_group.convert_to_stitch_sampler(dataloader.sampler)
>>> preds = {"recording_1": torch.rand(10, 1), "recording_2": torch.rand(20, 1)}
>>> targets = {"recording_1": torch.rand(10, 1), "recording_2": torch.rand(20, 1)}
>>> timestamps = {"recording_1": torch.rand(10, 1), "recording_2": torch.rand(20, 1)}
>>> metric_group.update(preds, targets, timestamps)
>>> metric_group.compute()
```

---
Notes:
- `MetricGroup` is too simple of a name, because we are still assuming that we have `recording_id > task_name > unit_name`, so not sure what we can call it. `CustomMetricGroup` is the best I can come up with currently.
- We can choose to separate `MetricGroup(...)` and `MetricGroup(..., enable_stiching=True)`  into two classes. the second could more explicitly include stitching in its name.
- Do we also want to separate things by depth? It's a bit much because we would have 6 different classes total, but might be worth it if we come up with a very clear naming system.

mazabou:
The reason why I like this is because if you already know the torchmetrics api, you're going to immediately get what's happening. nothing is "hidden", at least before introducing stitching (which is a good amount of magic / while being pretty straight forward from the user side).

re this comment: "here the symmetry between the metrics dict and targets/preds break because metrics has depth 3 but targets/preds has depth 2..."

we can solve this and keep the symmetry by  introducing a `MetricList` class. it can have a list of multiple metrics! and now the symmetry is preserved! 
```
>>> metrics = {
...     "recording_1": {
...         "task_1": MetricList( torchmetrics.Accuracy(), torchmetrics.F1Score()),
...         "task_2": torchmetrics.F1Score()
...     },
...     "recording_2": {
...         "task_1": torchmetrics.Accuracy(),
...     }
... }
>>> metric_group = MetricGroup(metrics)
>>> preds = {
...     "recording_1": {
...         "task_1": torch.rand(10, 1),
...         "task_2": torch.rand(20, 1)
...     },
...     "recording_2": {
...         "task_1": torch.rand(10, 1),
...     }
... }
>>> targets = {
...     "recording_1": {
...         "task_1": torch.rand(10),
...         "task_2": torch.rand(20),
...     },
...     "recording_2": {
...         "task_1": torch.rand(10),
...     }
... }
>>> metric_group.update(preds, targets)
>>> metric_group.compute()
```

You can even use the `MetricList` in depth 1, if you only are working with a single task, and don't want to specify a task_name.
```
>>> metric_group = MetricGroup({"recording_1": MetricList( torchmetrics.Accuracy(), torchmetrics.F1Score()), "recording_2": torchmetrics.R2Score(),})
>>> preds = {"recording_1": torch.rand(10, 1), "recording_2": torch.rand(20, 1)}
>>> targets = {"recording_1": torch.rand(10, 1), "recording_2": torch.rand(20, 1)}
>>> metric_group.update(preds, targets)
>>> metric_group.compute()
```

so you can do all these combos:
`recording_id`
`recording_id > metric_name` 
`recording_id > task_name`
`recording_id > task_name > metric_name`

vinamarora8:
I like the direction overall. But there are a few missing components that needs to be created for it to fit. 

The API currently implemented in the PR does is designed for 2 things:
1. Mimicing the API of `torchmetrics`, like you pointed out.
2. Being able to take in the (ugly) batched outputs of `prepare_for_readout` and `prepare_for_multitask_readout`.

Your suggestion does not solve (2). But I am sure that conversion of output format can be handled gracefully somehow and then the API you're suggesting will be a great fit. 
I also think there is more code needed for _creating_ a MetricGroup object, as the dataset config dict is not well aligned with how you propose instantiating this class. Although, this conversion detail can be hidden inside the `Datamodule`s.

Overall, I think the API you're suggesting is a great overhaul, and the MetricGroup would be a very reusable class. I am hugely in favor of giving users the ability to create their own evaluators easily.

However, if it will take more than a couple days to implement, I would propose merging this PR into `main`, as it does temporarily fix the unreasonable lightning callback magic of `main`; something I don't want to stay in `main` for too long.

---

Regarding naming the metric group class, I will throw in a suggestion: `NestedMetrics`, `NestedStitchedMetrics`.

mazabou:
Glad you like it. I agree regarding (2) it needs to be handled outside. I can make a PR into this branch tonight! If I am not able to make it, let's merge this tomorrow morning.

AlexandreAndr:
@vinamarora8  is this up to date or can we close

vinamarora8:
This is out of scope at the moment

vinamarora8:
@AlexandreAndr If I were to implement this again, it would look very different and much simpler (similar to how it is in examples/poyo now). So, let's attempt it once again in a while, after we do the merge!

PR #53, opened 2025-02-13
Fixed StitcherEvaluator bug and added [minimal] tests

milosobral:
The `on_validation_epoch_start` method always used the the `val_sequence_index` from the `trainer.data_module` which causes issues especially if no validation was ever done before attempting to _test_.

edited typo

### DIFF
--- a/tests/test_stitcher_evaluator.py
+++ b/tests/test_stitcher_evaluator.py
@@ -1,0 +1,69 @@
+import pytest
+import torch
+import lightning as L
+import torchmetrics
+from collections import defaultdict
+
+from torch_brain.utils.stitcher import MultiTaskDecodingStitchEvaluator
+
+
+class MockDataModule:
+    def __init__(self, sequence_index):
+        self.val_sequence_index = sequence_index
+        self.test_sequence_index = sequence_index
+
+
+class MockTrainer:
+    def __init__(self, sequence_index):
+        self.datamodule = MockDataModule(sequence_index)
+        self.loggers = []
+        self.is_global_zero = True
+
+
+@pytest.fixture
+def mock_metrics():
+    return {
+        "session1": {
+            "task1": {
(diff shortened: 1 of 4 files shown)


tests/test_stitcher_evaluator.py
vinamarora8, line 1: Thanks for adding this test! A small suggestion to make it a little bit better. For `test_on_validation_epoch_start` and `test_on_test_epoch_start`:
1. Set different `val_sequence_index` and `test_sequence_index` inside the MockDataModule (instead of both being `sequence_index`). 
2. Then ensure the correct sequence index was picked up by the StitchEvaluator in each of the two tests.

torch_brain/utils/stitcher.py
```diff
@@ -376,11 +357,38 @@ def on_validation_epoch_end(self, trainer, pl_module, prefix="val"):
...
+    def on_test_epoch_start(self, trainer, pl_module):
+        self._on_epoch_start_helper(trainer, mode="test")
 
     def on_test_batch_end(self, *args, **kwargs):
         self.on_validation_batch_end(*args, **kwargs)
 
     def on_test_epoch_end(self, *args, **kwargs):
         self.on_validation_epoch_end(*args, **kwargs, prefix="test")
+
+    def _on_epoch_start_helper(self, trainer, mode: str = "val"):
```
vinamarora8, line 369: Nit: Can we have a more explicit name for this function? Maybe `_setup_caches(...)`?

tests/test_stitcher_evaluator.py
vinamarora8, line 1: Can you please rename the test file to: `tests/test_multitask_decoding_stitch_evaluator.py`

vinamarora8 (changes requested):
Good catch! Changes look good overall, and thanks for adding some testing for the stitcher. I have left some comments for some small changes that should be made before we can merge this into main.

vinamarora8 (approved):
Awesome. Changes look good. Merging

PR #54, opened 2025-02-16
Add eval_transforms

mazabou:
We sometimes need to apply deterministic transforms on the data. Not sure what the best approach would be:
- Option 1: define train_transforms and eval_transformrs, the non-augmentation transforms will probably need to be repeated in each. (this is what I have implemented in this PR).
- Option 2: define train_transforms and transforms. where transforms would be applied to all splits, and train_transforms would be applied after the transforms.


vinamarora8:
I think Option 1 is good enough.

vinamarora8:
Can you please change the transform instantiation this way:
```python
eval_transforms = hydra.utils.instantiate(self.cfg.eval_transforms)
```
So that it's always required; but define the eval_transforms in the `base.yaml` config to be an empty list.
I don't like `cfg.get(...)` too much.

examples/poyo/train.py
```diff
@@ -159,19 +159,21 @@ def setup_dataset_and_link_model(self, model: POYO):
 
         self._init_model_vocab(model)
 
+        eval_transforms = hydra.utils.instantiate(self.cfg.get.eval_transforms)
+
```
vinamarora8, line 163: Bug. Should be:
```python
eval_transforms = hydra.utils.instantiate(self.cfg.eval_transforms)
```

vinamarora8 (changes requested):
There's are bug here. I am unable to train poyo with this. 

Let's make sure training works every time a PR is submitted for merging.

mazabou:
Edits made. Confirmed to be running. We need to have the auto-testing for training asap, let me know if you are able to do it soon, or I'll do it myself.

vinamarora8 (approved):

PR #55, opened 2025-02-17
Simplify stitching sampler logic

mazabou:
Currently the custom stitching sampler is used for three reasons:
1. it makes sure that samples that are meant to be stitched are sampled close to each other. 
2. it tracks the "groups" of samples that are meant to be stitched together, to know when they are fully processed, and that the cache can be flushed.
3. it makes sure that the groups of samples are sent to the same rank to avoid sending copies of the data across ranks.

There is a lot of back and forth between the dataset and the stitchevaluator, and currently this will only work for one sampler. 

This PR proposes a `convert_to_stitching_sampler` function. that can be applied to any sampler. This removes the need for the `DistributedStitchingSampler` and simplified the interface, making the use of the stitcher optional.

Fundamentally, the stitching sampler is not supposed to change the behavior of the sampler, so this makes sense. We can change the stitcher to make the call to the `convert_to_stitching_sampler` optional, and only recommend it if there is OOM error, or in the documentation.

### DIFF
--- a/torch_brain/data/sampler.py
+++ b/torch_brain/data/sampler.py
@@ -2,6 +2,7 @@ import math
 import logging
 from typing import List, Dict, Tuple, Optional, TypeVar, Iterator
 from functools import cached_property
+from collections import defaultdict
 
 import torch
 import torch.distributed as dist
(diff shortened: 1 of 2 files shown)


mazabou:
Replaced by #71

PR #57, opened 2025-02-17
Cleanup compute_loss_or_metric

mazabou:
We have already gotten rid of the "metric" part of `compute_loss_or_metric`. This allows us to define metrics outside of this giant function. Similarly, we can simplify the loss part and make it easier to edit by the user.

This PR defines the `Loss` class and implements `MSELoss` and `CrossEntropyLoss`. Any `Loss` module can be defined outside of `torch_brain` and attached to a modality.

Open to feedback.


torch_brain/nn/multitask_readout.py
```diff
@@ -1,26 +1,29 @@
...
 from torch_brain.data.collate import collate, chain, track_batch
-from torch_brain.registry import ModalitySpec, MODALITIY_REGISTRY
 from torch_brain.utils import (
     resolve_weights_based_on_interval_membership,
     isin_interval,
 )
 
+if TYPE_CHECKING:
+    from torch_brain.registry import ModalitySpec
+
```
mazabou, line 18: Note that this is needed because of the circular import issues

torch_brain/nn/loss.py
```diff
@@ -1,94 +1,112 @@
...
-    if output_type == DataType.CONTINUOUS:
-        if loss_or_metric == "mse":
-            # TODO mse could be used as a loss or as a metric. Currently it fails when
-            # called as a metric
-            # MSE loss
-            loss_noreduce = F.mse_loss(output, target, reduction="none").mean(dim=1)
+from torchtyping import TensorType
+
+
+class Loss(torch.nn.Module, ABC):
```
vinamarora8, line 9: Can we make this like pytorch. They name their loss abstract class with an underscore. `_Loss`. https://github.com/pytorch/pytorch/blob/1eba9b3aa3c43f86f4a2c807ac8e12c4a7767340/torch/nn/modules/loss.py#L38

mazabou: Our Loss class is not made to be private, it can be instantiated. Not sure why we would use `_Loss`

vinamarora8 (changes requested):
This is good. I am convinced we need to move towards custom loss classes.

Suggestion for PR: Would be cleaner if you move `torch_brain.nn` doc updates to a separate PR.

mazabou:
Yep, usually I try to stick to making edits on the docs corresponding to the edits in the PR. Will keep it more separate in the future.

@vinamarora8 what are the requested changes?

torch_brain/nn/loss.py
vinamarora8: Didn't think of that aspect. My bad. Not using underscore is good then.

vinamarora8:
It was that one comment above. Agree with you, I was wrong to request renaming `Loss` to `_Loss` there.

vinamarora8 (approved):
Great change! Merging

PR #58, opened 2025-02-17
add sampler tutorial

mazabou:
This requires adding data samples in docs/_data. Main question is whether this is ok. Currently brainsets does not allow us to download a subset of the sessions. 

Each file is 10Mb. Not sure if this will cause bloat. It would be messy to have to remove these files from the entire history.


vinamarora8 (changes requested):
These docs are 🔥. 

Re: Including data files. I would prefer if we avoid putting data files in git history. 
I think [Git LFS](https://git-lfs.com/) can be used here. It's a smart way to track "large" asset files that either don't change over time, or change by complete replacement (i.e. you don't care about their granular line-level diff, which is the case here)

mazabou:
I have added the files to an s3 bucket, and created a function that will download the files from it.
I also added readthedocs testing when PRs are created. The documentation build is successful which means we're good to merge, here's the link: https://torch-brain--58.org.readthedocs.build/en/58/concepts/sampler.html

vinamarora8 (commented):
Looks good!

vinamarora8 (approved):

PR #59, opened 2025-02-17
bug: handle case where rank does not have any samples from a given recording

mazabou:
Found a bug when doing distributed training. It happens when a rank ends up with no sample from a given recording. which leaves to an empty cache and thus an error.

The only solution is to create a tensor.empty, but requires knowing the dimensions of the prediction and targets. which can be infered from the modality_spec. 

- [x] Update DecodingStitchEvaluator
- [ ] Update multi task stitcher


mazabou:
Merging this fix until we finish https://github.com/neuro-galaxy/torch_brain/pull/71

PR #60, opened 2025-02-18
Cleanup configs

mazabou:


examples/poyo/configs/model/poyo_1.3M.yaml
```diff
@@ -1,4 +1,4 @@
-_target_: torch_brain.models.POYOPlus
+_target_: torch_brain.models.POYO
```
vinamarora8, line 1: nice catch

examples/poyo/train.py
```diff
@@ -287,10 +285,11 @@ def main(cfg: DictConfig):
         )
 
     # get modality details
-    readout_spec = MODALITIY_REGISTRY[cfg.readout_id]
+    readout_id = cfg.dataset[0].config.readout.readout_id
+    readout_spec = MODALITIY_REGISTRY[readout_id]
```
vinamarora8, line 289: Can you add a TODO comment that we need to make sure the readout_ids are the same for all datasets.

vinamarora8 (changes requested):
Comments added. There is a bug in both `defaults.yaml`.

Additional bug that stops poyo+ training: Change `base.yaml` to `default.yaml` in all `poyo_plus/configs/train_*.yaml`

mazabou:
Suggested edits made.

vinamarora8 (approved):
lgtm! merging

PR #62, opened 2025-02-18
Run POYO/POYO+ training in the testing workflow

mazabou:
Update the workflow to include a rule for downloading mc_maze_small, training POYO for 2 epochs, then training POYO+ for 2 epochs (including validation and testing). 

This PR includes adding the configuration files for mc_maze_small.

---

installing `brainsets` is done through cloning the main branch, but in the future (`brainsets 0.1.2` release) we will replace this with a `pip install`


vinamarora8 (approved):
Looks good! Very happy we have an end-to-end training CI test :D Thanks for adding it

PR #64, opened 2025-02-19
CI Testing: Speed up testing with `uv`

vinamarora8:
Currently Github Actions take ~3m45s to complete, out of which ~2m30s are spent just installing python packages with `pip install`. 

This PR brings in using [`uv`](https://pypi.org/project/uv/) for installing python packages. It's very fast and installs packages in parallel (raw `pip` does it serially). Using this reduces the total package installation time to ~45s, and overall testing time to ~2m.

| Task                 | Time before | Time with `uv` |
|----------------------|-------------|----------------|
| Install dependencies | 1m 34s      | 25s            |
| Install plugin       | 1s          | 0s             |
| Install brainsets    | 53s         | 19s            |
| **Total**            | **2m 28s**      | **44s**            |

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -22,17 +22,18 @@ jobs:
       run: |
         python -m venv venv
         source venv/bin/activate
+        pip install uv
     - name: Install dependencies
       run: |
         source venv/bin/activate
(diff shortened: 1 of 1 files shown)


mazabou (approved):
neat!

PR #67, opened 2025-02-20
Fix session ID assertions in test_dataset_sim.py for consistency

AlexandreAndr:
Following the change of this PR #40.
Is targeted for local tests


vinamarora8 (approved):
looks good

PR #69, opened 2025-02-21
Fix CI tests: Silent error in `brainsets prepare`

vinamarora8:


mazabou (approved):
Weird, I don't have an issue when running it locally so this must be a workflow problem. We will eventually go back to using pip install brainsets for version 0.1.2 which will have pei_pandarinath_nlb_2012

PR #71, opened 2025-02-24
Introduce MetricGroupWithStitcher

mazabou:
This follows the conversation in #52 ([comment](https://github.com/neuro-galaxy/torch_brain/pull/52#issuecomment-2676543586)). The idea is to embrace the torchmetrics API a bit more, and remove the model-specific definition to make the stitcher and evaluation more general. Landed with the name `MetricGroupWithStitcher`, there would also be a `MetricGroup` that we can define in the future that will not do any stitching. 

This also incorporates the suggested change to the stitcher, making things cleaner and more transparent (#55).

`MetricGroupWithStitcher` can be single task or multi task by simply passing in a tensor or a dict of tensors. It also supports having multiple metrics computed for the same data.

POYO is very clean and readable imo, but POYO+ has to do a lot of work to adapt its tokens (this is specific to the MultiTaskReadout module, so we can eventually absorbe this in a function).

The code runs for both POYO and POYO+. I want to add a bit more tests, and validate that everything is correct before merging. As well as add documentation, and examples. I am opening this PR in case there is feedback for changing aspects of this. 

A few pain points:
- how do we handle device in general?
- when returning the computed metrics, the dict is auto-formated to be `{recording_id}/{metric_name} or `{recording_id}/{task_name}/{metric_name}` is that okay?


mazabou:
Note that the subpackage `torch_brain.metrics` was added because we are going to start having neuro-specific metrics that we will start adding for models like NDT/NEDS...

torch_brain/metrics/metric_group.py
vinamarora8, line 1: This is a lot of new code, and I appreciate that. But I don't understand two decisions here:

1. Support for only depth == 1 or 2. 
  I liked your original idea about supporting an arbitrary depth and arbitrary structure of metrics. What's stopping us from doing that?
2. `recording_id` in the argument of `update()`. 
  Again, I liked your original idea better where inputs (`timestamps`, `preds`, `targets`) were structured in the format of the metric dictionary. That was very general. Having `recording_id` in the argument list makes it way less general of a class. Right now, it is more like a "MultiRecordingMetricGroup" iykwim. Anything stopping us from implementing what you proposed before?

mazabou: I tried to implemented in the most general way, but ended up finding challenges:
1. because of the caching mechanism, we know that each time `update` is called, we will get the results from a given sample, defined by (recording_id, start, end). we also are tracking the sample_idx in the sampler, to know when it's time to flush the cache. The way the cache tracker is incremented is based on that first layer (depth=1) so it automatically implies that depth=1 needs to always be recording

https://github.com/neuro-galaxy/torch_brain/blob/cd34047bdccf9bc5123040ed94e23cb51a265bf5/torch_brain/metrics/metric_group.py#L196-L200

2. the last depth needs to always be metric, this is because the metrics will be applied on the same data (so we don't need a different cache for each metric). 

https://github.com/neuro-galaxy/torch_brain/blob/cd34047bdccf9bc5123040ed94e23cb51a265bf5/torch_brain/metrics/metric_group.py#L189-L192

3. This leaves one file layer, which is task. each task will have its own tokens, so it needs to be cached differently (hence why it needs to be treated differently from the metric layer). it is also different from the recording layer, because a new task doesn't increment the sampler sequence_index, only a new sample from a trial can. 

I have tried thinking of other use cases that cannot be covered by `recording > task > metric` that would warrant supporting a depth > 2. Also currently we support:
```
recording > metric
recording > multiple metrics
recording > task > metric
recording > task > multiple metrics
```

this basically leaves the following which we can implement if you think is needed
```
task > metric / multiple metrics
```
any other use case we are not covering or a way we can think of making this work? 
I realized that we cannot define arbitrary dictionaries because each layer implies a certain way of dealing with things. 

---

regarding `recording_id` being used in `update()`, you are right. that was the plan. Initially I wanted to support adding a batch in `update` but then realized that multiple samples in a batch could have the same recording_id, which meant that I would mess up the order of the samples (and thus break sequence_id).

I made the change such that `update()` takes in one sample at a time.

i can make this change:
```
metric_group.update(preds, targets, timestamps, recording_id)
```
it would be:
```
metric_group.update({recording_id: preds}, {recording_id: targets}, {recording_id: timestamps})
```

it's a bit redundant, since here we are updating one sample at a time.

I also have the option of making `update` accept batches, but in this case we have to keep the list structure, and cannot use the same dict because we will have multiple samples with the same key (recording_id)

---

finally regarding how this MetricGroup is super specific. I guess we can name it something else, but fundamentally, for this package, we already are calling our dataset, `Dataset` and not `MultiRecordingDataset`. I am happy to support this use case if you think its needed:
```
task > metric / multiple metrics
```

PR #74, opened 2025-03-03
Bugfix: Inverted mask in `xformers` rotary attention forward

vinamarora8:
The mask to bias transformation in `rotary_attn_xformers_func` is currently the opposite of what it should be.
The bug only surfaces if you have `xformers` installed in your environment. If you're using pytorch attention, you would be unaffected.

### DIFF
--- a/torch_brain/nn/rotary_attention.py
+++ b/torch_brain/nn/rotary_attention.py
@@ -80,7 +80,7 @@ class RotaryCrossAttention(nn.Module):
             - x_context: (B, N_c, D_c)
             - query_pos_emb: (B, N_q, D_h)
             - context_pos_emb: (B, N_c, D_h)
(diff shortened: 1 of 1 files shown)


mazabou (approved):
Thanks @vinam for fixing this. Did you try to run with xformers to confirm that the outcome would be the exact same as the one with pytorch? I am also happy to test this if not.

vinamarora8:
> Thanks @vinam for fixing this. Did you try to run with xformers to confirm that the outcome would be the exact same as the one with pytorch? I am also happy to test this if not.

Just tested ~50 epochs on 2 sessions of MP. It's not _exactly_ the same; I would attribute it to different numerical precision in the two kernels. However, it atleast trains with this fix. It wasn't learning at all when the mask was flipped (train loss wouldn't go down, as you would expect)

image

PR #83, opened 2025-03-12
add POYO-1 configs

mazabou:


vinamarora8:
Would be good to finalize and merge #49 before this

PR #85, opened 2025-03-12
replace setup.py with pyproject.toml

mazabou:
@vinamarora8 could you give me the email you would like to use? 
re: dependencies, we want to relax some of the versions? 
I think we can remove: lightning, torch-optimizer, wandb and tabulate (we would need to add torch-metrics)
then we add a requirements.txt in POYO, or we make them optional-dependancies?


vinamarora8:
- Can you use `vinam@posteo.de` for my email.
- We should definitely remove lightning, wandb, etc. and put it in `requirements.txt` in `example/*/`. Not as optional dependencies.

pyproject.toml
```diff
@@ -0,0 +1,46 @@
...
+classifiers = [
+    "Programming Language :: Python :: 3",
+    "License :: OSI Approved :: MIT License", 
+    "Operating System :: OS Independent",
+]
+requires-python = ">=3.8"
+dependencies = [
+    "temporaldata",
+    "numpy<2.0.0",
+    "torch>=2.0.0",
```
vinamarora8, line 24: Can we make these:
```
numpy~=1.0
torch~=2.0
```
That indicates to the package managers that we're okay with any `numpy` and `torch` with major-versions 1 and 2 respectively.

PR #86, opened 2025-03-12
Support datasets without a "config"

vinamarora8:
Currently, POYO forward pass is broken if I define dataset like
```python
dataset = Dataset(
    root="../data/processed", 
    recording_id="perich_miller_population_2019/c_20151106_center_out_reaching",
)
```
because the dataset lacks the "config" fields that are only possible with a yaml-based dataset config.

This breakage stems from `prepare_for_readout` expecting "config.readout" to be present. However, since `prepare_for_readout` already has access to `readout_spec`, it should not _require_ the presence of a readout config.

This PR fixes the problem.

### DIFF
--- a/torch_brain/utils/readout.py
+++ b/torch_brain/utils/readout.py
@@ -20,6 +20,9 @@ def prepare_for_readout(
     readout_config = dict(readout_spec)
 
     if "readout" in data.config:
+        # Ensure readout config contains valid keys and merge it with the
+        # default config provided by readout_spec
+
         _readout_config = data.config["readout"]
 
         required_keys = ["readout_id"]
--- a/torch_brain/utils/readout.py
+++ b/torch_brain/utils/readout.py
@@ -18,7 +18,7 @@ def prepare_for_readout(
 ):
 
     readout_config = dict(readout_spec)
-    
+
     if "readout" in data.config:
         _readout_config = data.config["readout"]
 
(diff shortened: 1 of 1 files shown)


torch_brain/utils/readout.py
```diff
@@ -16,40 +16,45 @@ def prepare_for_readout(
...
+
+    readout_config = dict(readout_spec)
+
+    if "readout" in data.config:
+        # Ensure readout config contains valid keys and merge it with the
+        # default config provided by readout_spec
+
+        _readout_config = data.config["readout"]
+
+        required_keys = ["readout_id"]
```
vinamarora8, line 28: Note: I am aware that `readout_id` is completely redundant here (because `readout_spec` is a parameter of this function). I'm keeping it here, just so that the readout config format of `prepare_for_readout` and `prepare_for_multitask_readout` remain similar.

torch_brain/utils/readout.py
```diff
@@ -16,40 +16,45 @@ def prepare_for_readout(
...
-        )
-
-    key = readout_config["readout_id"]
-
-    if key not in torch_brain.MODALITY_REGISTRY:
-        raise ValueError(
-            f"Readout {key} not found in modality registry, please register it "
-            "using torch_brain.register_modality()"
-        )
-
```
vinamarora8, line 50: Removed this, `key` isn't needed anywhere.

PR #88, opened 2025-03-21
add option for not leaving files open in Dataset

mazabou:


vinamarora8:
Heyy, what's the use-case for this? Is it to handle "too many open files"?

torch_brain/data/dataset.py
```diff
@@ -308,6 +312,13 @@ def get(self, recording_id: str, start: float, end: float):
 
         return sample
 
+    def _get_data_object(self, recording_id: str):
+        if self.keep_files_open:
+            return copy.copy(self._data_objects[recording_id])
+        else:
+            file = h5py.File(self.recording_dict[recording_id]["filename"], "r")
+            return Data.from_hdf5(file, lazy=True)
+
```
vinamarora8, line 321: I think we need to figure out a way to close these file handles. Otherwise you're opening a new handle every sample, and it remains dangling after use (right?). Could create OS level problems.

You _could_ populate and make use of the `_open_files` dictionary. The files in this dictionary are taken care of by the `_close_open_files()` method during object deletion.

vinamarora8 (changes requested):
It looks good. The functionality you're adding needs to be present. Just worried about too many open files, look at my comment below.

PR #89, opened 2025-03-24
Adding 2d padding in collater

AlexandreAndr:
# Padding Variable-Sized 2D Tensors Within a Batch

Imagine having a tensor with 2D shapes that vary within the batch and across dimensions. We would like to pad each tensor at the bottom and the right to ensure consistent sizes.

# Use Case
A typical use case is when creating a causal attention mask. This mask can be more complex than a simple triangular matrix because the timestamps can been shuffled within the batch (e.g. NDT2 and MAE based models).

Some examples can be found in the test file.

# Current Implementation
The current implementation might not be optimal, but I have an intuition that preallocating memory is still a good practice.

One alternative approach is to use `torch.pad` for each tensor using a for-loop and then stack them. However, based on my experience, this tends to be suboptimal compared to preallocating memory.

I can try benchmarking both methods to provide empirical results. Additionally, I’m very interested if you have suggestions for a more optimal solution.

# Possible Optimizations
We could consider implementing the 2d padding function at a lower level in C++ using JIT, but I am unsure if it's worth the effort, especially if the function is used in only a few models.

### DIFF
--- a/torch_brain/data/collate.py
+++ b/torch_brain/data/collate.py
@@ -2,11 +2,10 @@ import copy
 from collections import namedtuple
 from typing import Callable, Dict, List, Optional, Tuple, Type, Union
 
-import torch
-from torch.utils.data._utils.collate import collate as _collate, default_collate_fn_map
-
 import numpy as np
-
+import torch
+from torch.utils.data._utils.collate import collate as _collate
+from torch.utils.data._utils.collate import default_collate_fn_map
 
 # pad
 PaddedObject = namedtuple("PaddedObject", ["obj"])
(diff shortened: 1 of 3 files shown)


torch_brain/data/collate.py
```diff
@@ -110,6 +109,57 @@ def pad8_collate_object_fn(
...
+
+
+def track_mask2d(input: Union[torch.Tensor, np.ndarray]):
+    r"""Wrap an array or tensor to specify that its padding mask should be tracked. This
+    is used in conjunction with :obj:`pad2d`.
+
+    Args:
+        input: An array or tensor.
+    """
+    assert input.ndim == 2
```
vinamarora8, line 131: Would be better to raise error `if input.ndim != 2` instead of having an assertion here.

torch_brain/data/collate.py
```diff
@@ -110,6 +109,57 @@ def pad8_collate_object_fn(
...
+    return pad2d(torch.ones_like(input, dtype=torch.bool))
+
+
+def pad2d_collate_tensor_fn(
+    batch,
+    *,
+    collate_fn_map: Optional[Dict[Union[Type, Tuple[Type, ...]], Callable]] = None,
+):
+    assert all(elem.ndim >= 2 for elem in batch)
```
vinamarora8, line 140: Same here: raising an error would be preferred over assertion.

vinamarora8:
I think supporting 2D padding is good; will come in handy for many things in the future, beyond masks. 
Regarding optimizing the implementation with a C++ backend: Agreed that it does not seem necessary at the time, and can be deferred to later when python implementation seems to be a bottleneck.

AlexandreAndr:
I updated them! Do you think it is the good way to do it?

PR #90, opened 2025-03-25
Add `reinitialize_vocab` in `InfiniteVocabEmbedding`

vinamarora8:
When finetuning, we have to re-init the vocab for which we currently follow 2 steps: `extend_vocab` and then `subset_vocab`. I have received feedback that the two steps look weird, and having a single method for the purpose of reinitializing would make the code's intent easier to understand.


torch_brain/nn/infinite_vocab_embedding.py
```diff
@@ -109,6 +109,34 @@ def initialize_vocab(self, vocab: List[str]):
 
         self.initialize_parameters(len(self.vocab))
 
+    def reinitialize_vocab(self, vocab: List[str]):
+        r"""Reinitialize the vocabulary with a new list of words. This method will clear
+        the existing vocabulary and embeddings, and initialize a new vocabulary.
+
+        Args:
+            vocab (List[str]): A list of words to initialize the new vocabulary.
```
AlexandreAndr, line 117: rm `(List[str])`

AlexandreAndr (commented):
I think this is a great feature for the users I was not a huge fan of expand and subset workflow. I think we could merge this as it is? 
Just one minor comment on the docstring

AlexandreAndr:
cc @vinamarora8

AlexandreAndr:
Also update the changelog and potentially some tutorials if we have one with finetuning?

vinamarora8:
I'm not 100% of the API atm (mainly the name, and if we just absorb this in initialize_vocab itself).

AlexandreAndr:
I see yes maybe `initialize_vocab` with a flag my be good

PR #93, opened 2025-04-09
Add resampling and pre_split_transform feature

AlexandreAndr:
In some cases, we need to have the decoding behavior at a specific sample rate and require resampling (either upsampling or downsampling) the data. Currently, our framework does not support resampling.

To address this, we introduce a new feature: `pre_split_transform`.

This feature allows applying a transformation to the data before it is sliced, as opposed to the standard transform, which operates on already-sliced data. This is particularly useful for resampling, as it enables us to process a larger window than the final slice size. By doing so, we can better mitigate edge artifacts commonly introduced during resampling operations.

### DIFF
--- a/torch_brain/transforms/resampling.py
+++ b/torch_brain/transforms/resampling.py
@@ -14,18 +14,13 @@ class Resampler:
     ):
         self.base_frequency = base_frequency
         self.resample_frequency = resample_frequency
-        # TODO add the ref from where this come from
+        # TODO add a ref from where this comes from
         if extra_window_factor < 4 or 10 < extra_window_factor:
             raise ValueError("extra_window_factor must be between 4 and 10, inclusive.")
         self.extra_window_factor = extra_window_factor
-        # TODO not sure of this check
-        if self.base_frequency / self.resample_frequency % 1 != 0:
-            raise ValueError(
-                "The base frequency must be a multiple of the resample frequency."
-            )
-        self.extra_window_length = (
-            extra_window_factor * self.base_frequency / self.resample_frequency
-        )
+
+        b_f, r_f = self.base_frequency, self.resample_frequency
+        self.extra_window_length = extra_window_factor * max(b_f / r_f, r_f / b_f)
 
     def __call__(self, data: Data, start: float, end: float):
         return self.resample(data, start, end)
@@ -42,8 +37,7 @@ class Resampler:
         data = copy.deepcopy(data)
 
         # TODO: intersting in you opinion about this
-        # When we are in a center window, because we divide by two the extra_window_length we could
-        # be benefical to ensure we are using an offset to keep start and end in the sampled point in the new window
+        # This design ensure us to keep start as the first up/down sampled point
         offsett = self.extra_window_length / 2 % self.resample_frequency
         if (
             data.domain.start <= start - self.extra_window_length / 2 - offsett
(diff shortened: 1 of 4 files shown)


torch_brain/transforms/resampling.py
```diff
@@ -0,0 +1,65 @@
...
+
+        Args:
+            data (Data): The data to resample.
+            start (float): The start time of the new interval.
+            end (float): The end time of the new interval.
+        """
+        # TODO center the interval around the start and end time
+        out = data.__class__.__new__(data.__class__)
+        data = copy.deepcopy(data)
+        new_data = data.slice(start, end + self.extra_window_length)
```
mazabou, line 28: here we would want to use `reset_origin=False`

mazabou: and use start - self.extra_window_length

torch_brain/transforms/resampling.py
```diff
@@ -0,0 +1,65 @@
...
+        for key, value in new_data.__dict__.items():
```
mazabou, line 29: should we make the attributes to resample a list here? instead of resampling everything

mazabou (commented):
This makes sense overall, I know you are still working on this, I left a few comments

torch_brain/transforms/resampling.py
AlexandreAndr: New implementation will center the window in priority

torch_brain/transforms/resampling.py
AlexandreAndr: I'm not sure I understand the nuance here

AlexandreAndr:
The pipeline is failing due to an error when slicing `RegularTimeSeries` see this [fix](https://github.com/neuro-galaxy/temporaldata/pull/39).
What should I do wait for the bug fix to be release on temporaldata or comment these tests?

AlexandreAndr:
Also I am not super proud of the test cases. I can refactor them to test both regular/irregular in a big data object.
Or even add stronger cases. Do you have suggestions on that? I am out of inspiration

torch_brain/transforms/resampling.py
```diff
@@ -0,0 +1,108 @@
...
+        Args:
+            data (Data): The data to resample.
+            start (float): The start time of the new interval.
+            end (float): The end time of the new interval.
+        """
+        out = data.__class__.__new__(data.__class__)
+        data = copy.deepcopy(data)
+
+        # TODO: intersting in you opinion about this
+        # This design ensure us to keep start as the first up/down sampled point
```
AlexandreAndr, line 40: Need your opinion here

torch_brain/transforms/resampling.py
```diff
@@ -0,0 +1,108 @@
...
+                    resampled_time_based_values[timekey] = x_out
+
+                unsliced_value = IrregularTimeSeries(
+                    timestamps=t_out,
+                    **resampled_time_based_values,
+                    timekeys=val._timekeys,
+                    domain="auto",
+                )
+
+                # TODO the slicing will be process after anyway, here it could be ommited for enficency purpose
```
AlexandreAndr, line 82: Need your opinion here

AlexandreAndr:
If you want to review the resample method as discussed I tag you here @vinamarora8

AlexandreAndr:
Resampling is complicated need dependencies on scipy. I feel it is out of scope at the moment

PR #94, opened 2025-04-14
resampling

mazabou:


torch_brain/transforms/resampling.py
```diff
@@ -1,29 +1,32 @@
...
-        base_frequency: float,
-        resample_frequency: float,
-        extra_window_factor: int = 10,
+        target_sampling_rate: float,
+        target_keys: List[str],
     ):
-        self.base_frequency = base_frequency
-        self.resample_frequency = resample_frequency
-        # TODO add a ref from where this comes from
-        if extra_window_factor < 4 or 10 < extra_window_factor:
```
AlexandreAndr, line 18: So no need to enforce it to be in [4,10]?

torch_brain/transforms/resampling.py
```diff
@@ -33,76 +36,69 @@ def resample(self, data: Data, start: float, end: float):
...
+                x_in = getattr(timeseries, attr_key)
+                x_out = decimate(x_in, downsample_factor, axis=0, ftype="iir")
+                setattr(timeseries, attr_key, x_out)
+
+            timeseries.sampling_rate = self.target_sampling_rate
+
+        return data
+
+
+def irregular_to_regular_timeseries(
```
AlexandreAndr, line 69: Do we want that here and not inside the `IrregularTimeSeries` object? 
We could just call it as `ts.to_regular()`?

mazabou: Yes, I was thinking we can iterate on it here, and once we are satisfied, we can add it directly to temporaldata once it's finalized

torch_brain/transforms/resampling.py
```diff
@@ -1,29 +1,32 @@
...
 
-        b_f, r_f = self.base_frequency, self.resample_frequency
-        self.extra_window_length = extra_window_factor * max(b_f / r_f, r_f / b_f)
+        # To avoid aliasing, we add a buffer of 10 * sampling_rate / target_sampling_rate
+        # points, which is equivalent to 10 / target_sampling_rate seconds
+        self._anti_aliasing_buffer = 10.0 / self.target_sampling_rate
 
     def __call__(self, data: Data, start: float, end: float):
-        return self.resample(data, start, end)
+        start = start - self._anti_aliasing_buffer
```
AlexandreAndr, line 24: Doing it like that makes the buffer of 20 * sampling_rate / target_sampling_rate no?

PR #95, opened 2025-04-25
Update package name from pytorch_brain to pytorch-brain

AlexandreAndr:


mazabou (commented):
The name of the package is correct `pytorch_brain`, pip changes `_` to `-` but they are the same, and running `pip install pytorch_brain` works.

since in python, we are calling `torch_brain` I think it's better to keep it consistent in the documentation and call it `pytorch_brain`.

PR #96, opened 2025-05-23
Added tests (and removed unused imports)

milosobral:
Verifying that the output from the varlen version of the forward pass is the same as the output from the normal version.

### DIFF
--- a/torch_brain/nn/rotary_attention.py
+++ b/torch_brain/nn/rotary_attention.py
@@ -71,6 +71,10 @@ class RotaryCrossAttention(nn.Module):
         self.to_kv = nn.Linear(context_dim, inner_dim * 2, bias=False)
         self.to_out = nn.Linear(inner_dim, dim)
 
+        self.use_efficient = os.getenv(
+            "TORCH_BRAIN_USE_EFFICIENT_ATTENTION", "true"
+        ).lower() in ("true", "1", "yes")
+
     def forward(
         self,
         x_query,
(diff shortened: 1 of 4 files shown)


mazabou:
Not sure how to best approach this. What do you think about making it an environment variable?

vinamarora8:
Can the env variable be read once during the `__init__` method? Something about reading env vars every forward pass doesn't feel right.

vinamarora8:
Totally fine with changing min py version to 3.9!

torch_brain/nn/rotary_attention.py
```diff
@@ -65,6 +71,10 @@ def __init__(
         self.to_kv = nn.Linear(context_dim, inner_dim * 2, bias=False)
         self.to_out = nn.Linear(inner_dim, dim)
 
+        self.use_efficient = os.getenv(
+            "TORCH_BRAIN_USE_EFFICIENT_ATTENTION", "true"
+        ).lower() in ("true", "1", "yes")
+
```
vinamarora8, line 77: I'm a little bothered by the vagueness of the phrase "efficient". Perhaps:
```suggestion
        self.use_xformers = os.getenv(
            "TORCH_BRAIN_USE_XFORMERS", "true"
        ).lower() in ("true", "1", "yes")
```

torch_brain/nn/rotary_attention.py
```diff
@@ -467,6 +486,7 @@ def rotary_attn_xformers_func(
         value,
         attn_bias=attn_bias,
         p=dropout_p,
+        op=(xops.fmha.cutlass.FwOp, xops.fmha.cutlass.BwOp),
     )
```
vinamarora8, line 490: This definitely feels like bad practice. We should find another approach so that we don't set the op like this. I think slightly older version of xformers did not have this problem.

If we have to resolve to fixing the op, then we should atleast set it to flash when bf16 is detected.

AlexandreAndr:
@milosobral is it still up to date or should we close?

milosobral:
This was a workaround for a bug that existed in a specific version of xformers which was fixed so closing this

PR #98, opened 2025-06-04
SequentialFixedWindowSampler: Remove assert requiring `step_size <= window_size`

vinamarora8:
The assert `step_size <= window_size` is overly restrictive. I've been in situations where I needed `step_size > window_size`.


mazabou (approved):
I am ok with this.
Curious to understand what the use case would be?

PR #99, opened 2025-07-10
Fixed Device bug in infinite_vocab_embedding

milosobral:
This would move the embedding layer back to the CPU.


nandahkrishna (approved):
Makes sense to me. Thanks!

PR #101, opened 2025-08-04
Milo add value mapping mtreadout

milosobral:
Added a Value mapping in the registry to make it easy to define remaps of certain values directly from the config files. 

This is useful  when doing multiclass classification where not all classes are used. For example, if I have classes "right_hand":0, "left_hand":1 and "feet":2, and I want to classify right_hand vs feet, This PR would allow me to easily remap those two classes to IDs 0 and 1 and ignore left_hand by defining:
```
value_map: {0: 0, 1: -1, 2: 1}
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -162,3 +162,5 @@ data/raw
 **/.DS_Store
 
 checkpoint.pth
+
+uv.lock


vinamarora8 (commented):
Hey Milo, can this be solved with a Transform instead? Just have a transform that would do the value map in the Data slice itself.

The reason I'm saying this is because I think this is getting a bit too deep into the registry based configuration. If there are other good mechanism to achieve the same can effect, I would prefer that.

vinamarora8:
You're not wrong. I understand the thought that modality registry should be a place to add everything about the modality.

But I'm worried about this - wouldn't this open the doors for a lot of bloat in the registry? So many small variations of tasks are possible if we start subsetting class labels. Where will we draw the line?
(apologies if i'm being too paranoid)

PR #102, opened 2025-08-04
Add uv lockfile to the gitignore

milosobral:
See discussion on discord thread


pyproject.toml
```diff
@@ -8,16 +8,16 @@ version = "0.1.0"
...
 ]
-license = {text = "Apache-2.0"}
+license = { text = "Apache-2.0" }
 classifiers = [
     "Programming Language :: Python :: 3",
     "License :: OSI Approved :: Apache Software License",
     "Operating System :: OS Independent",
 ]
-requires-python = ">=3.8"
+requires-python = ">=3.9"
```
vinamarora8, line 20: Although I don't think anyone uses py<3.9 amongst the devs, this should still be a separate PR, no?

vinamarora8 (commented):
Hey @milosobral, sorry about the delay in reviewing this PR! Just mentioned a small thing to discuss below

pyproject.toml
milosobral: Maybe I added it here because when building an environment with uv it is impossible to use python 3.8 because of some dependency conflict so I thought I would fix it while I'm at it. If I rename this PR to also include that change would you be okay with that?

vinamarora8: I think if you build the env without `uv`, it'll work with py3.8. Atleast it worked last time I tested.
I don't think we should change it over here. 
I personally don't use < py3.10, but I feel we should have a discussion with rest of the users before changing the minimum required version. If no one is using < py3.10, then I would much prefer just switching to 3.10 directly.

PR #104, opened 2025-10-03
add SelectByRegion and RandomSelectByRegion augmentations

mazabou:
This adds two new transforms, that both assume there is a `data.units.region` attribute. 

Currently both support providing a single region, but in the future which support a group of regions.


torch_brain/transforms/select_by_region.py
```diff
@@ -0,0 +1,151 @@
...
+        region: str = None,
+        reset_index: bool = True,
+    ):
+        self.field = field
+        self.reset_index = reset_index
+        self.region = region
+
+    def __call__(self, data: Data):
+        # get regions from data
+        regions = data.units.region
```
vinamarora8, line 34: I think this can be a parameter of the class, with this default value. 
`region_field = "units.region"`. Sometimes datasets will use different names to denote region.

torch_brain/transforms/select_by_region.py
```diff
@@ -0,0 +1,151 @@
...
+        self,
+        field: str = "spikes",
+        exclude_regions: Optional[List[str]] = None,
+        min_units: int = 1,
+        reset_index: bool = True,
+        seed: Optional[int] = None,
+    ):
+        self.field = field
+        self.reset_index = reset_index
+        self.exclude_regions = exclude_regions if exclude_regions is not None else []
```
vinamarora8, line 99: I think can be replaced with simpler:
```python
self.exclude_regions = exclude_regions or []
```

torch_brain/transforms/select_by_region.py
```diff
@@ -0,0 +1,151 @@
...
+    def __call__(self, data: Data):
+        # get regions from data
+        regions = data.units.region
+        unique_regions, region_counts = np.unique(regions, return_counts=True)
+
+        # filter out excluded regions and regions with insufficient units
+        available_regions = []
+        for region, count in zip(unique_regions, region_counts):
+            if region not in self.exclude_regions and count >= self.min_units:
+                available_regions.append(region)
```
vinamarora8, line 112: Can be optimized with some numpy ops:
```python
available_regions = unique_regions[
    (region_counts >= self.min_units) 
    & ~np.isin(unique_regions, exclude_regions)
]
```

torch_brain/transforms/select_by_region.py
```diff
@@ -0,0 +1,151 @@
...
+        region (str): Region to select.
+        reset_index (bool, optional): Whether to reset the index of the units. Set this to False
+            if you want to set the activity of the dropped units to 0, and to True if you
+            want to reindex the units to create a new sub-population. Defaults to True.
+    """
+
+    def __init__(
+        self,
+        field: str = "spikes",
+        region: str = None,
```
vinamarora8, line 25: `region` should not have a default value of `None` IMO. It should be a required parameter.

vinamarora8 (changes requested):
Good PR! I have left some suggestions below.

Less related: Reminds me of a cleanup we should do soon-ish. We should abstract out the unit masking logic into a util function. UnitDropout and the two transforms you're adding here, all share the same code for applying a unit mask.

torch_brain/transforms/select_by_region.py
mazabou: Initially, I tried generated a generic SelectBy transform but here's the challenge: we need to know that `units` is tied to `spikes` somehow, because neurons selected based on `units` will be filtered in `spikes` and that connection is missing from all our classes currently. we need a way of formalizing that link so that more of these operations can be done correctly. 

this is why I reverted back to SelectByRegion specifically (Region = region) and we can figure out how we want to generalize this more broadly to more attributes. Any thoughts?

torch_brain/transforms/select_by_region.py
vinamarora8: why do you want field to come first?

torch_brain/transforms/select_by_region.py
vinamarora8: Okay, I understand this. But maybe we at least allow the phrase "region" being flexible? I am okay with "units" being fixed

torch_brain/transforms/select_by_region.py
mazabou: good question. mentally, you first select what the target object is and then what region you want to select in that object. 

```
SelectByRegion(field="spikes", region="CA3")
```
versus.
```
SelectByRegion(region="CA3", field="spikes")
```

this might be more of a nit, so I am happy to switch the order

milosobral (commented):
I feel like there is a lot of repeated code between SelectByRegion and RandomSelectByRegion. Maybe the Random version should subclass SelectByRegion and simply add the logic to select the random region where it is relevant?

Also, might be a stupid question but is there a reason why you wouldn't support selecting multiple regions at the same time? Is that not a use case that is relevant?

torch_brain/transforms/select_by_region.py
vinamarora8: Hmm, IMO we should just have `region` first if `field` has a default value and `region` does not. 
BUT I'm not too stuck on it. I'm okay if you want to keep it this way, and then _maybe_ add a check for region not being `None`?

vinamarora8:
Hey @mazabou, planning to get this in soon?

AlexandreAndr:
Hey @mazabou, @vinamarora8 @milosobral 
`SelectByRegion` might be a special case of `UnitFilterByAttr` implemented here #171 (for the attribute region)?
Maybe we could add another transform like `RandomUnitFilterByAttr`?
I am also fine to have more specific/special cases of tansform like `SelectByRegion` but we could benefit from wraping them on something more general like `UnitFilterByAttr`. What do you think?

vinamarora8:
I would also prefer something like `UnitFilterByAttr` to do the job of `SelectByRegion`. We don't really have a fixed convention yet for where is "region" stored and what it is called. With using `UnitFilterByAttr`, we'll never have to rely on having such a convention (even though someday we might have one).

Re: `RandomSelectByRegion`: I'm not sure if it should be provided by us. Doesn't seem that generally useful. And users can write it easily (once we provide some utilities to mask out unit easily)

PR #105, opened 2025-10-03
add SkipOnFailure

mazabou:
Useful transform container that can be applied to any augmentation that might fail, and that would be skipped.
This is likely to happen when working with a lot of datasets.


vinamarora8 (commented):
This will be a good addition. We can merge this after #104 is in.
Also, adding a test would be nice.

milosobral (dismissed):
Agreed about adding a test, otherwise LGTM

AlexandreAndr:
@mazabou @milosobral @vinamarora8 I asign this to myself, its a cool feature and almost ready to go (need the tests)

torch_brain/transforms/container.py
```diff
@@ -83,3 +85,47 @@ def __call__(self, data: temporaldata.Data) -> temporaldata.Data:
...
+        if self.backup_copy:
+            data_backup = copy.deepcopy(data)
+
+        try:
+            return self.transform(data)
+        except Exception as e:
+            logging.warning(
+                f"Transform {self.transform} failed. The following exception was raised: {e}\n"
+                f"Restoring pre-transform data."
+            )
```
vinamarora8: I don't think emitting a warning should be default behavior. By using this transform, the user is explicitly asking us to not care about transform failures, and we should do it silently. (I'm usually not a fan of warning emissions in the regular code-flow path, as you can tell 😅 )

I do see that such a warning would help in debugging. So this could be toggled-on by a constructor flag (e.g. `warn: bool = False`)

vinamarora8 (commented):
See my comment above!

torch_brain/transforms/container.py
AlexandreAndr: I agree!

torch_brain/transforms/container.py
```diff
@@ -83,3 +87,59 @@ def __call__(self, data: temporaldata.Data) -> temporaldata.Data:
...
+        transform: transformation to attempt to apply.
+        backup_copy: whether to make a backup copy of the input data. Set to True if the
+            transform risks mutating the input data in-place before raising an exception.
+            Defaults to False.
+        warn: whether to emit a warning when the transform fails. Defaults to False.
+    """
+
+    def __init__(
+        self, transform: Callable, backup_copy: bool = False, warn: bool = False
+    ) -> None:
```
vinamarora8, line 119: I think 
1. `backup_copy` should be `True` by default. Safer defaults are generally better than unsafe defaults, atleast in this context I would say (small perf impact + only few users will use this)
2. With this ^, you won't need to emit warning based on `backup_copy`, since the user is aware of what they're doing

vinamarora8, line 119:
```suggestion
    def __init__(
        self, transform: Callable, backup_copy: bool = False, warn: bool = False
    ):
```
contructors shouldn't have a return type

torch_brain/transforms/container.py
```diff
@@ -83,3 +87,59 @@ def __call__(self, data: temporaldata.Data) -> temporaldata.Data:
...
+        ...             raise ValueError("not enough units")
+        ...         return data
+        >>> transform = SkipOnFailure(RequireMinUnits(min_units=10))
+
+    Args:
+        transform: transformation to attempt to apply.
+        backup_copy: whether to make a backup copy of the input data. Set to True if the
+            transform risks mutating the input data in-place before raising an exception.
+            Defaults to False.
+        warn: whether to emit a warning when the transform fails. Defaults to False.
```
vinamarora8, line 114: Example below args

Issue #111, opened 2025-11-06
New methods to simplify usage for inference tasks


luiztauffer:
I’ve been thinking on ways to make it as user-friendly as possible the task of:
1. loading an existing Brainset
2. loading a pre-trained POYO model
3. running inference on the Brainset using the pre-trained POYO model. 

Here are some possibilities:

### 1. POYO.tokenize method
The `tokenize` method is too strict, hardcoding data attributes names, e.g. `spike_timestamps = data.spikes.timestamps`. We should parametrize the data-dependent variables, and this can be done, for example, at a POYO method that loads the data, see POYO.set_data_loaders method below.

### 2. POYO.load_pretrained method
Should we have a new class method load_pretrained? That aims at facilitating the loading from checkpoint files. E.g.:

```python
@classmethod
def load_pretrained(checkpoint_path):
    checkpoint = torch.load(ckpt_path, map_location="cpu", weights_only=False)
    model = POYO(**checkpoint["hyper_parameters"]['model'])
    state_dict = {k.replace("model.", ""): v for k, v in checkpoint["state_dict"].items()}
    model.load_state_dict(state_dict)
    return model
```

### 3. POYO.set_data_loaders method
Looking at the functions defined in the cosyne_tutorial_part_2.ipynb notebook, it feels like some of these could be general enough to become a POYO method? The `get_train_val_loaders` for example, could become `POYO. set_data_loaders`, which receives as input a Brainset path or config, and automatizes for the user the tasks of:
- creating train, val and test DataLoaders based on the target Brainset
- reinitialize the vocabs for the target Brainset
- connect tokenizers to Datasets
- overall, couple this instantiation of POYO with a target Brainset

I’m happy to break these ideas into separate Issues and make tentative PRs for each one, but I’d love to hear your opinion on them first.

luiztauffer:
@vinamarora8 thanks for your comments!

### POYO.load_pretrained

good, then I'll start there!

### POYO.set_data_loaders

If it turns out to be general enough, we could eventually have a parent class with general methods from which all torch_brain model inherit from, and with abstractmethods indicating where specific models should implement their own code.

> how will that method know what the train_transforms are

one optional argument of `POYO.set_data_loaders` could be `transformation_fnc: Callable`, and users could pass their custom transformation function if their Brainset is special enough. 

I think having these functionalities controlled by POYO (or any other torch_brain model) will allow us e.g. to run validation of data types and shapes at any point we want, and return informative errors.

[...]

PR #112, opened 2025-11-08
POYO.load_pretrained method

luiztauffer:
Load pretrained method for POYO

see discussion: https://github.com/neuro-galaxy/torch_brain/issues/111

- [x] depends on https://github.com/neuro-galaxy/torch_brain/pull/114

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -41,7 +41,11 @@ jobs:
       run: |
         source venv/bin/activate
         export WANDB_MODE=dryrun
-        pytest tests/
+        pytest --cov=torch_brain --cov-report=xml --cov-report=term-missing -v
+    - name: Upload coverage reports to Codecov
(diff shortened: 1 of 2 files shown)


vinamarora8 (changes requested):
Hey, the feature looks good. But I think we should also add a test for this. 
In the test you could save a random init POYO model, and then load it with your new classmethod.

vinamarora8:
Actually, I just remembered, we have checkpoint hosted on an S3 bucket. You can access it with (taken from the COSYNE25 tutorial):

```python
import boto3
from botocore import UNSIGNED
from botocore.config import Config

# Download the pretrained model
s3 = boto3.client('s3', config=Config(signature_version=UNSIGNED))
s3.download_file(
    "torch-brain",
    "model-zoo/poyo_mp.ckpt",
    "poyo_mp.ckpt"
)
```

PR #113, opened 2025-11-08
Proposal - Dataset config schema definition for POYO

luiztauffer:
This is a proposal to systematize the way in which we define Dataset configurations schema for POYO, using Pydantic models.

**Problem:**
I noticed that  `data.config`:
- defines data-related configuration parameters specific to a given model (e.g. readout vs multitask_readout)
- some of these parameters are required (readout_id) and some are optional (weights)
- values filled are specific to that Brainset

I’m discovering how to use it by trial and error, and suggest a more systematic approach:
- torch_brain models should explicitly define the schema for their dataset configs. This will allow for validation not only of presence of keys, but also types of values, and provide better documentation on these fields
- since this is model-dependent, I believe this should be implemented either as a classmethod or with pydantic models for each torch_brain model
- we can then perform proper validation when calling a model

Ps.: if we move forward with this, we would probably want to think on how to better organize these schemas within the project. The current placement is just a first idea, but probably insufficient, since other model-related schemas might also be necessary.

### DIFF
--- a/tests/test_poyo_dataset_schema.py
+++ b/tests/test_poyo_dataset_schema.py
@@ -20,22 +20,22 @@ def test_validate_real_dataset_configs(config_path):
     # Load config
     cfg = OmegaConf.load(config_path)
     config_dict = OmegaConf.to_container(cfg, resolve=True)
-    
+
     # Validate - should not raise ValidationError
     validated = POYODatasetConfig.model_validate(config_dict)
-    
+
     # Basic assertions
     assert len(validated) > 0, "Config should have at least one selection block"
-    
+
     # Check first block has required structure
     first_block = validated[0]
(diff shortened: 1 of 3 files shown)


milosobral (commented):
100% agree with the idea of adding a clearer way to enforce proper configuration.
In order to actually enforce this, what if we also had a `TorchBrainModel` abstract class that enforces the `DatasetConfig --> tokenize --> forward` pipeline and documents it properly?

luiztauffer:
the implementation of the schema here is independent of that possible new class abstraction, what do you think of this proposed way, using Pydantic? @milosobral

vinamarora8:
Closing since dataset configs are deprecated

PR #114, opened 2025-11-08
Update required Python minimal version to 3.10

luiztauffer:
Python 3.9 just reached end of life, meaning there's no more official support for that version (which is quite old by now): https://devguide.python.org/versions/

I suggest we update the minimal required Python version to 3.10

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
-dev = ["pytest", "black==24.2.0", "pre-commit>=3.5.0", "flake8"]
+dev = ["pytest<9", "black==24.2.0", "pre-commit>=3.5.0", "flake8"]
 
 [project.urls]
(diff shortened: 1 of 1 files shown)


docs/source/concepts/installation.rst
```diff
@@ -1,7 +1,7 @@
 Installation
 ============
 
-**torch_brain** is available for Python 3.9.
+**torch_brain** is available for Python > 3.10.
```
vinamarora8: Same as above

README.md
```diff
@@ -29,7 +29,7 @@
 
 
 ## Installation
-torch_brain is available for Python 3.9 to Python 3.11
+torch_brain is available for Python > 3.10 and can be installed via pip.
```
luiztauffer:
```suggestion
torch_brain is available for Python >= 3.10 and can be installed via pip.
```

vinamarora8 (approved):
lgtm

PR #117, opened 2025-11-11
conda env create file

luiztauffer:
utility file for easy creating a conda environment for torch_brain


luiztauffer:
this is simply a convenience file to reproduce a working environment, if they use conda (I do)

PR #118, opened 2025-11-11
TorchBrainModel base class and Dataset configuration

luiztauffer:
**WIP - this is preliminary attempt, and open to discussion and changes!**

This PR brings a couple of significant changes:
- A base class `TorchBrainModel` to be inherited by all torch_brain models. 
- New methods for the parent and child classes
- Pydantic classes defining Dataset configurations
- Helper functions at `utils/training.py`
- [An example](https://github.com/neuro-galaxy/torch_brain/blob/torchbrainmodel/examples/poyo/finetune.py) of how much easier it will be for a new user to use a pre-trained model with their own Brainset, after these changes (compare it to the Cosyne tutorial)

## Base class
The current base class is inspired on POYO and what I could extrapolate from it. I do expect us to change it as we go looking at other models and observe other needs for abstraction. For example, @vinamarora8  raised a valid point that not every torch_model will be a decoding model. A potential extension then would be to have `TorchBrainModel` with more basic methods and e.g. `DecodeTorchBrainModel` which inherits from it with readout_spec related operations. Examples:

- `TorchBrainModel` -> `DecodeTorchBrainModel` -> `POYO`
- `TorchBrainModel` -> `EmbeddingTorchBrainModel` -> `MyEmbModel`

I think that this first round already illustrates how useful such abstractions can be. See e.g. the `set_datasets` method in [parent](https://github.com/neuro-galaxy/torch_brain/blob/878f3d72e0d7e73efb6c466c16f08e0a5069416e/torch_brain/models/base_class.py#L79-L111) and [child](https://github.com/neuro-galaxy/torch_brain/blob/878f3d72e0d7e73efb6c466c16f08e0a5069416e/torch_brain/models/poyo.py#L374-L383) classes.

## New methods for the parent and child classes
Implementing some operations as modular methods allows torch_brain to:
- be user friendly, with minimal usage configuration using sensible defaults
- be flexible enough to allow override of any specific methods

Example of things that can be automatized / enforced:
- given a dataset configuration, the model class should be able to know it will bind it's own tokenize method to that dataset, and that it should reinitialize vocab, [see here](https://github.com/neuro-galaxy/torch_brain/blob/878f3d72e0d7e73efb6c466c16f08e0a5069416e/torch_brain/models/poyo.py#L374-L383)
- given a model hyperparameter configuration, the data sampler `window_length` should match the model `sequence_length`, [see here](https://github.com/neuro-galaxy/torch_brain/blob/878f3d72e0d7e73efb6c466c16f08e0a5069416e/torch_brain/models/poyo.py#L392-L398)

Again, those a sensible defaults, and it is very simple to override them for more advanced users.

## Dataset configurations as Pydantic classes
[See them here](https://github.com/neuro-galaxy/torch_brain/blob/torchbrainmodel/torch_brain/schemas/base_class/dataset_schema.py). The original PR for that is [here](https://github.com/neuro-galaxy/torch_brain/pull/113), but I brought them to this one so the models methods can make proper use of them.

This will help standardize the usage and documentations of the dataset configuration across the project. This is necessary for:
- Easing the creation and understanding of such configurations by the users
- Validation and quick error detection
- Automatize some operations across models methods and helper functions

Pydantic classes are composable, that means we could follow a similar pattern of inheritance from parent configs to child configs, and each model can define it’s own needs while respecting a basic common structure.

Right now, users still need to manually set values for `readout_spec` and `dataset_config`, which feels unnecessary, since `dataset_config` seems to contain `readout_spec`. This is something we can discuss how to improve, but right now you can see that we can already validate basic stuff, e.g. make sure they contain the same values, [see here](https://github.com/neuro-galaxy/torch_brain/blob/878f3d72e0d7e73efb6c466c16f08e0a5069416e/torch_brain/models/base_class.py#L56-L77).

Since each model might need different `dataset_config`, they can also implement helper methods that output a minimal valid config, including (optionally) calculating statistics necessary for that specific model, [see here](https://github.com/neuro-galaxy/torch_brain/blob/878f3d72e0d7e73efb6c466c16f08e0a5069416e/torch_brain/models/base_class.py#L146-L207)

## Helper functions
I copied them from the Cosyne tutorials, I brought the basic necessary ones for finetuning, and haven’t put much work on generalizing or improving them. Something to work on!

### DIFF
--- a/examples/poyo/finetune.py
+++ b/examples/poyo/finetune.py
@@ -1,0 +1,61 @@
+import torch
+import torch_brain
+from torch_brain.models import POYO
+from torch_brain.registry import ModalitySpec, DataType
+
+
+# User defines readout properties
+readout_modality_name = "wheel"
+readout_value_key = "wheel.vel"
+readout_timestamp_key = "wheel.timestamps"
+
+# Helper class method to create minimal dataset config 
+dataset_config = POYO.create_basic_dataset_config(
+    dir_path=".",
+    brainset="ibl",
+    sessions=["sub-CSH-ZAD-024_ses-8207abc6-6b23-4762-92b4-82e05bed5143-processed-only_behavior"],
+    readout_id=readout_modality_name,
+    value_key=readout_value_key,
+    timestamp_key=readout_timestamp_key,
+)
+
+# Define readout specification
+# PS: we could probably automatize this using the dataset config
+readout_spec = ModalitySpec(
+    id=100,
+    name=readout_modality_name,
+    value_key=readout_value_key,
+    timestamp_key=readout_timestamp_key,
+    dim=1,
+    type=DataType.CONTINUOUS,
+    loss_fn=torch_brain.nn.loss.MSELoss(),
+)
+
+# Helper class method to load pre-trained model checkpoint
+model = POYO.load_pretrained(
+    checkpoint_path="poyo_mp.ckpt",
+    readout_spec=readout_spec,
+    skip_readout=True,
+)
+
+# Load model to device
+# PS: we could probably automatically do this (by default)
+device = (
+    torch.device("mps") if torch.backends.mps.is_available()
+    else torch.device("cuda:0") if torch.cuda.is_available()
+    else torch.device("cpu")
+)
+model.to(device)
+
+# Helper method to bind dataset to model
+model.set_datasets(
+    dir_path=".",
+    dataset_config=dataset_config,
+)
+
+# Helper method finetune model with provided Brainset
+poyo_ft_r2_log, poyo_ft_loss_log, poyo_ft_train_outputs = model.finetune(
+    num_epochs=2,
+    epoch_to_unfreeze=10,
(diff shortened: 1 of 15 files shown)


milosobral (commented):
Sorry it took me so long to get to this!

I think this PR maybe is doing a bit too much at once. 
 * I like the basic idea of having a TorchBrainModel Superclass that enforces the presence of some functions (load_pretrained, tokenize...). However, I don't think that having a finetune method is great when it comes to separation of concerns. I think its better if the models in torch_brain/models are completely independent from the training code.
 * I also think the Dataset configuration is a mess right now and needs to be reworked but this should maybe be its own PR.
 * I agree that models and datasets are coupled as the tokenize function of each model is used as a transform of the other. As you suggested, maybe the better way to see things is that Transforms should be applied from the model side? Not 100% convinced either way here.

vinamarora8:
Closing since we have moved to a config-free Dataset class

PR #123, opened 2025-11-14
[BUILDATHON][TEAM PIXEL-PEOPLE] Adds plain Attention layer (without rotary)

vinamarora8:
The existing `nn.RotaryCrossAttention` and `nn.RotarySelfAttention` have API and logic for when rotary embeddings are used. This PR adds a non-rotary `nn.CrossAttention` and `nn.SelfAttention`.

Xformers is probably still buggy, same as Rotary versions of this; this is an existing problem that should be solved in a separate PR.

### DIFF
--- a/torch_brain/nn/attention.py
+++ b/torch_brain/nn/attention.py
@@ -41,8 +41,8 @@ class CrossAttention(nn.Module):
         *,
         dim: int,
         context_dim: Optional[int] = None,
-        heads: int = 8,
-        dim_head: int = 64,
+        heads: int,
+        dim_head: int,
         dropout: float = 0.0,
     ):
         super().__init__()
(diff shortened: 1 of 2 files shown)


torch_brain/nn/attention.py
```diff
@@ -0,0 +1,453 @@
...
+            num_heads=self.heads,
+            dropout_p=self.dropout if self.training else 0,
+            attn_mask=context_mask,
+        )
+
+        # project back to dim
+        out = self.to_out(out)
+        return out
+
+    def forward_varlen(self, x_query, x_context, query_seqlen, context_seqlen):
```
milosobral: We probably want to add the type hints as well (although to be fair they are not there in rotary_attention.py either)

milosobral (commented):
Some small comments but other than that looks good. 

There is a lot of overlap between this and the rotary attention classes which I don't love but I couldn't come up with a good way of mitigating that that doesn't add some unreasonable level of complexity

PR #125, opened 2025-11-14
[BUILDATHON] [TEAM LAYER V] Patching Transform

milosobral:
Added a patching transform for RegularTimeSeries.

Wasn't sure what the best implementation would be for IrregularTimeSeries (or if there even is one or if it's very use case specific). I can add if people think it's useful.


AlexandreAndr:
@milosobral how far are we to have this ready?
I can help for the irregular, it is used on NDT2 for example

milosobral:
> @milosobral how far are we to have this ready?
I can help for the irregular, it is used on NDT2 for example

@AlexandreAndr Just like with masking, I kind of have changed my mind and tend to think that we should probably leave it to each model to define what patching means as part of the tokenization. If we see that a lot of models use very similar patching, we can think about adding an implementation as a transform. For now, this feels like it would just be very specific to what I think patching should be like.

vinamarora8:
I agree with @milosobral. Let's wait for a common use-case to appear in code before creating an abstraction for it. This generally leads to better API design, even if it is obvious right now that such an abstraction might be useful.

PR #127, opened 2025-11-14
[BUILDATHON] [TEAM FALCON] Adds augmentations

divyansha1115:


AlexandreAndr:
@vinamarora8 @milosobral @divyansha1115 I feel those augmentations are cool and standard, do we want to merge them? Might testing/cleaning

divyansha1115:
@AlexandreAndr I think these are only compatible with IrregTS right now, but it should be fairly straightforward to make them work with RegTS as well. Cleanup and testing is for sure needed.

It’d also be great to add a practical use-case for these, maybe in the tutorials/docs, showing how they can be used/chained together.

AlexandreAndr:
I agree you for the tutorials/docs. I also saw that you did cool ones here: #138 and #140. Maybe we can group them or the opposite split them and select the one we would like to prioritize. What do you think?

divyansha1115:
I think those are same as or very close to the ones added in this PR. I think @vinamarora8 told me to split them up at some point which is why they are separate. Some of these like RateDrift might not be super useful. I saw it being used in just one paper. We can identify the ones that are most common in literature and prioritize those.

vinamarora8:
So the consensus is to split these up? I would prefer that too. Ideally, we would want to have transforms to be general and applicable to data with either IrregTS or RegTS signals. But ofcourse, sometimes it is impossible

PR #128, opened 2025-11-14
[BUILDATHON] [TEAM PIXEL-PEOPLE] Add same session batch sampler

lpjiang97:
Add a batch sampler that return batches with samples from the same session. Supports shuffling within sessions and across sessions. Test under `tests/test_sampler/test_same_session_batch_sampler`.

Minor fixes:
- Typehints and docstrings in `data/sampler.py`: all `sample_intervals` are now type `Interval`
- formats


AlexandreAndr:
@vinamarora8 @milosobral This can be a quite useful feature, I had to use it and reimplement when I was implementing NDT/MtM for instance. There have been some changes on the samplers file organization so I can help @lpjiang97 if we decide to go for it

lpjiang97:
Let me know if this is still needed and I can fix the merge conflicts

AlexandreAndr:
Hey @lpjiang97, Yes that would be super usefull!
A first step would be to move the sampler from `torch_brain/data/sampler.py` to its own file here `torch_brain/samplers/`. 
I will alsp do a first pass review

torch_brain/data/sampler.py
```diff
@@ -273,6 +273,102 @@ def __iter__(self):
             yield from indices
 
 
+class SameSessionBatchSampler(torch.utils.data.Sampler):
+    r"""Randomly samples batches of trials from given trial intervals. Each batch contains
+    trials from the same recording session.
+
+    Args:]
```
AlexandreAndr, line 280: `Args:]` -> `Args:`

torch_brain/data/sampler.py
```diff
@@ -273,6 +273,102 @@ def __iter__(self):
...
+        sampling_intervals (Dict[str, Interval]): Sampling intervals for each
```
AlexandreAndr, line 281: `sampling_intervals (Dict[str, Interval]):` -> `sampling_intervals:`
The doc automatically generates the type so no more need to provide them with `(type)`
The same applies to the other args bellow

torch_brain/data/sampler.py
```diff
@@ -273,6 +273,102 @@ def __iter__(self):
...
+            every recording session.
+
+    """
+
+    def __init__(
+        self,
+        *,
+        sampling_intervals: Dict[str, Interval],
+        batch_size: int,
+        generator: Optional[torch.Generator] = None,
```
AlexandreAndr, line 301: `Optional[torch.Generator]` -> `torch.Generator | None`
Optional is not the most recent convention to do this typing

torch_brain/data/sampler.py
```diff
@@ -273,6 +273,102 @@ def __iter__(self):
...
+    r"""Randomly samples batches of trials from given trial intervals. Each batch contains
+    trials from the same recording session.
+
+    Args:]
+        sampling_intervals (Dict[str, Interval]): Sampling intervals for each
+            session in the dataset.
+        generator (Optional[torch.Generator], optional): Generator for shuffling.
+            Defaults to None.
+        batch_size (int): Number of trials to sample per batch.
+        shuffle (bool, optional): Whether to shuffle the indices with a recording session.
```
AlexandreAndr, line 286: `shuffle` is set to False by default
`shuffle_session`  is set to True by default
I am curious if you would have a motivation for these choices of default values?

torch_brain/data/sampler.py
```diff
@@ -273,6 +273,102 @@ def __iter__(self):
...
+class SameSessionBatchSampler(torch.utils.data.Sampler):
+    r"""Randomly samples batches of trials from given trial intervals. Each batch contains
+    trials from the same recording session.
+
+    Args:]
+        sampling_intervals (Dict[str, Interval]): Sampling intervals for each
+            session in the dataset.
+        generator (Optional[torch.Generator], optional): Generator for shuffling.
+            Defaults to None.
+        batch_size (int): Number of trials to sample per batch.
```
AlexandreAndr, line 285: `batch_size` should be above `generator` in the docstring

AlexandreAndr:
I have a proposal: what if we were making this sampler more general, a wrapper arround any sampler to force it to produce session with same session within a batch?

This could be the interface
```
class SessionBatchSampler(torch.utils.data.Sampler[List[DatasetIndex]]):
    """Wraps any sampler yielding :class:`DatasetIndex` and groups its output
    into per-session batches.

    All indices in a batch share the same session id. The inner sampler controls
    the order of indices within each session; ``shuffle_batches`` controls the
    order of batches across sessions.

    Args:
        sampler: Inner sampler whose ``__iter__`` yields :class:`DatasetIndex` objects.
        batch_size: Number of samples per batch.
        shuffle_batches: Whether to shuffle the final batch order. Defaults to False.
        generator: Generator used when shuffling batches. Defaults to None.
        drop_last: Whether to drop the last incomplete batch per session.
            Defaults to True.
    """

```

AlexandreAndr:
Note: `recording_id` attribute on a data object is what is used for a session so maybe we could modify the Sampler to `RecordingBatchSampler`

AlexandreAndr:
What do you think @lpjiang97 and @vinamarora8 ?

vinamarora8:
@AlexandreAndr I like the idea. Better than having a session-batch-xyz-sampler for every xyz

PR #132, opened 2025-11-14
[BUILDATHON] [TEAM LAYER V] basic masking transforms 

milosobral:
Masks all RegularTimeSeries Objects and IrregularTimeSeries objects where the timestamps are evenly spaced.
Each of the fields in the masked object should be of the same dimension in the channel dimension.
Also assumes that channels are always in the second dimension.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,6 +40,9 @@ dev = [
     "tqdm",
     "pytest-cov",
 ]
+lightning = [
+    "lightning>=2.5.6",
+]
 
 [project.urls]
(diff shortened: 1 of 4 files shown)


AlexandreAndr:
Hey @milosobral what would be the usecase of this transform?

milosobral:
@AlexandreAndr The idea was to unify how masking is done in a transform. However, after playing with this idea for a while, I think it wouldn't be super useful most of the time because it is probably too opinionated for most use cases. Better to leave it to each model to define how the masking should be done

PR #146, opened 2025-11-23
Versioning and Release Workflow

vinamarora8:
This PR automates the versioning and release workflow.  

**Main philosophy:** We want a single source of truth for version numbers to keep life simple and avoid mistakes. 
Here, I use git tags as that source of truth, which is a standard approach.

## Release Workflow:
We will mark a release using a git tag. Then, a GitHub Action (included in this PR) will build and upload the sdist and wheel to PyPI/testPyPI.

**Overall workflow:**
1. Ensure `main` is ready to release  
2. Make sure `CHANGELOG.md` is up to date (review commit history to verify)  
3. Create and push the release tag (launches the Github Action)
```bash
$ git tag v0.x.y
$ git push origin v0.x.y 
```
4. Add a Github Release with a small description of changes from the last release.

**PyPI vs TestPyPI:** 
If the tag is like
- `v0.1.2-rc.1`: this is a release _candidate_, and goes to TestPyPI
- `v0.1.2`: this is the main release, and goes to PyPI

We can restrict permissions so only certain people can push tags. 
And we can also add protections such that persons A,B,C have to approve before the dist actually goes to PyPI (after the tag is pushed).

**Note:** Any pushes/merges to the main branch will go to TestPyPI (acting like a bleeding-edge/nightly release). The version number for these "nightly" releases is described below.

## Versioning
To keep the version number in sync with the tag, here I use `setuptools-scm`, which reads version info directly from the git tag. (Docs for `setuptools-scm`: [here](https://setuptools-scm.readthedocs.io/en/latest/))

**Development Versions:**
Nothing extra to do: `setuptools-scm` automatically generates unique development version identifiers. If someone is, say, 5 commits past the v0.1.1 last release tag, their installed package would report:

```bash
$ python -c "import torch_brain; print(torch_brain.__version__)"
0.1.2.dev5+g9a8b7c
```
This makes it easy for bleeding-edge users to report issues on the `main` branch; and we would know the exact commit they are on.  

## TODO
- [x] Pushes/merges to main should release on TestPyPI

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -23,11 +23,13 @@ jobs:
         with:
           python-version: "3.10"
 
-      - name: Install pypa/build
-        run: python3 -m pip install build --user
+      - name: Install uv
+        uses: astral-sh/setup-uv@v4
+        with:
+          enable-cache: true
 
       - name: Build a binary wheel and a source tarball
-        run: python3 -m build
+        run: uv build
 
       - name: Store the distribution packages
         uses: actions/upload-artifact@v4


milosobral:
Hey, a few comments:
* I think you can build directly with uv like this. 
```
jobs:
  build:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
        with:
          fetch-depth: 0 # Required for hatch-vcs to see git tags

      - name: Install uv
        uses: astral-sh/setup-uv@v3
        with:
          enable-cache: true

      - name: Build package
        # uv will use the build-system defined in pyproject.toml (hatchling)
        run: uv build
``` 
If this is anything like installing dependencies, it might be faster than using pip but you might need to use a different build system (in this example, it uses something called hatchling).
* I was thinking that we could maybe release to testPyPI on every push to main so that if someone wants the bleeding edge latest version, it's easy to always get it from there. Is there any reason why we might not want that?
* I don't love how we do release notes in a separate Markdown file that has to be manually edited. It would be better to use the github features where release notes are alongside the release. I found this tool that you can integrate directly with releases called `release-drafter` which you can setup to automatically create release notes from pull requests and/or branch names. I think that would be pretty cool, it would just mean that we need to be more explicitly careful about being clear in the commit messages from each PR (which can probably be made easier by using some sort of template)

milosobral:
> Just skimmed `release-drafter`'s readme. It seems like a cool and useful integration. I would be supportive of that PR, if you want to take charge!

Yea I think we can maybe merge this one first and leave improvements for a separate PR

milosobral (dismissed):
sorry forgot I needed to go back and review this again to approve

vinamarora8:
@milosobral made it so that pushes/merges on main branch will lead to publishing on TestPyPI. Also now we build using `uv build`

milosobral (approved):
LGTM!

vinamarora8:
> Bleeding edge can always be pip installed using git + branch.

@milosobral what do you think about Luiz's point about letting bleeding edge installs be only through git?
testPyPI does seem like a place more for testing releases rather than serving releases to users.

PR #148, opened 2025-11-25
Fix: Resolve state_dict corruption in InfiniteVocabEmbedding (addressing TODO)

ModeEric:
## Summary
This PR fixes a bug where `load_state_dict` was modifying the source model's vocabulary in-place, leading to corruption during model reloading. It also re-enables the `test_state_dict_loading` test case which was previously failing due to this issue.

## Changes
-Changed the vocabulary saving mechanism to use a shallow copy (`self.vocab.copy()`) instead of a direct reference. This prevents `pop()` operations during the `_hook` phase from corrupting the original source model.
-Addresses the `TODO` left in the codebase regarding the state dictionary.
-Re-enabled `test_state_dict_loading` (verified passing) and added a missing `@pytest.mark.skipif` to `test_get_brainset_ids` for better test hygiene.

## Test Plan
- Ran `pytest tests/test_embeddings.py`, and it passed (previously failing).

### DIFF
--- a/tests/test_dataset_sim.py
+++ b/tests/test_dataset_sim.py
@@ -1,6 +1,5 @@
 import os
 import tempfile
-from pathlib import Path
 
 import h5py
 import numpy as np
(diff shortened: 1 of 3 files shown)


vinamarora8 (approved):
Changes too look good, thanks Eric!
Merging!

Issue #149, opened 2025-11-26
Dummy Data


ghazalehran:
Is it possible to simply copy-paste the Dummy data that Milo used for EEG in other torch_brain repo here? and possibly extend it to spikes dummy , caalcium dummy etc? This way we don't have to create a dummy data for every test file (less redundancy) @milosobral

PR #150, opened 2025-12-02
Added basic dummies and generator function

milosobral:
Adressing Issue #149 

It would be quite useful if each of the brainsets pipeline defined what the resulting Data object looks like. Before doing that, a good first step is to have evrything defined directly in torch_brain (where it would be the most useful) so I added my proposition for a Spec to define and generate what the data would look like. 

This will: 
 - Help users of the brainset by being able to see what they can expect to find in a Data object without having to run the pipeline and open the object.
 - Facilitate realistic testing of torch_brain by creating synthetic data that closely replicates the appearance of real objects.
Curious to hear what you guys think!

For now, I used an LLM to generate most of the actual specs given that I don't have much experience with the spiking data so I wanted to have your opinion on the overall idea before spending too much time making sure all the brainset definitions actually match.


ghazalehran:
First, I wanted to reflect on whether adding and maintaining a Spec per brainset justifies the development cost (following our discussion from yesterda). My conclusion is Yes :)) here’s why. 

Extra cost:
A pipeline author needs to write a small structured description (the Spec) of fields that the pipeline already implicitly defines. In practice this is maybe 10–20% more work than writing the pipeline alone.

Benefits (much larger):

1. Explicit contract between brainsets and Torch Brain

- No more guessing what fields a brainset returns.

- Easier debugging when pipelines evolve.

- The Spec becomes the source of truth for documentation.

2. High-quality dummy data “for free”

- generate_dummy_data(brainset_id) can produce realistic, multimodal mock Data objects.

- This prevents every test file from reinventing its own dummy EEG/spikes/calcium constructors.

- It drastically improves test coverage with minimal boilerplate.

- Value-wise, that’s easily 30–40% improvement in test maintainability.

3. Future extensibility

- Specs enable automatic validation (assert real_data matches Spec),

- automatic docs generation (fields, shapes, sampling rates),

- potential visualization/debugging tools.

- This is another 20%+ long-term benefit.

Overall, the tradeoff feels very favorable: a small upfront investment for 60–80% improvement in maintainability, clarity, and tooling around brainsets. (numbers are not exact of course :))) )

ghazalehran (commented):
I won’t nit-pick the code since I know this is an early, rough version and you’ll probably refine the implementation later. Instead I have a comment on the current code: This Spec system at the brainset level makes a lot of sense, and having a declarative description of each dataset’s Data structure will definitely help both users and downstream tooling.

I think it would be valuable to add a complementary, higher-level layer for modality/task-specific dummy data. e.g.:

make_dummy_eeg_classification_data(in_chans, in_times, n_classes, …)

make_dummy_spike_decoding_data(n_units, duration, …)

make_dummy_calcium_data(n_rois, n_frames, …)

This layer would sit on top of the brainset Specs (or even be partially generated from them), but would allow developers working at the model level (Torch Brain) to generate compact, controllable dummy data without tying their tests to a specific brainset.

This separation would give us:

brainset-level Specs: “What does the actual pipeline output look like?”

task/modality-level helpers: “Give me a minimal but realistic dummy Data object for EEG / spikes / calcium so I can test models and samplers.” That would let users control things like in_times, n_channels, etc., without needing to care which brainset they’re emulating.

I’d be happy to help integrate this once the base Spec layer stabilizes.

milosobral:
@ghazalehran 

> I think it would be valuable to add a complementary, higher-level layer for modality/task-specific dummy data. e.g.:

I'm not sure what this high level lyer would achieved that is ont already achieved by the brainset specific dummies. Why would you want to make sure your code "supports" a Data object that will likely never exist?

luiztauffer:
Here’s how I imagine this working best:
- `temporaldata` classes generating specs automatically. See a proposal for this here: https://github.com/neuro-galaxy/temporaldata/pull/55
- `brainset` uses this functionality and just adds an extra layer of metadata to it (and potentially the dummy generator default options)
- `torch_brain` carries the dummy generator machinery

Once we have specs that can be loaded directly from Brainsets themselves, I think a better implementation of a `generate_dummy_data()` function would be:
1. parse the Brainset (or a single temporaldata Data object) spec
2. generate dummy data for each component of that brainset / Data object

I’m not sure where `distribution` and `params` should go: either at `brainset` or `torch_brain.generate_dummy_data()`

milosobral:
@luiztauffer Not sure I understand how this would be integrated with the average workflow when adding a brainset. Would it be that the developper of the pipeline would run `spec = data.generate_spec()`  on one of the generated Data objects and then manually copy and paste it somehwere alongside their pipeline so it can be found by torch brain? What happens when there is a change, do we make it part of the "regular" process that you HAVE to run this somewhere to make sure the spec is up to date?  

The issue I have with the specs being stored directly in `brainsets` is that I don't really see a use for them in there, the only realistic use I can see so far is for the dummy data for tests in `torch_brain`. I like your idea in theory and I initially thought the same thing but I feel like adding that much code across three different repos might be a bit overkill.

PR #154, opened 2025-12-18
New `Dataset` class

vinamarora8:
This PR introduces the `torch_brain.dataset` module containing a new way to create PyTorch Datasets. The existing way overly relies on configuration files written in YAML, which has proven hard to use, debug, and document. Type checking, IDE jump-to-definition, and data exploration are almost impossible in a YAML-based approach.

With this new `Dataset` class, the power of creating dataset objects is back to Python; while moving towards features for which people have expressed desire: handling complex train/val/test splits, easy dataset exploration in notebooks, handling modality specificity of Dataset, ...

**TL/DR**
- A new base `Dataset` class handles recordings of _one_ brainset and is instantiated using a simple pure Python API.
- `NestedDataset` class allows composing multiple `Dataset` objects for multi-brainset training, and are infinitely nestable.
- `Dataset` is designed to be extensible: we can (and do) have brainset-specific `Dataset` subclasses with custom functionalities, such as splits management.
- `Mixins` are included to add common modality-specific methods to the base `Dataset` class.

**Note**: If you prefer using dataset configs, things will not change _too_ much for you. Look at the "Concerns" section at the end.

## Dataset: `torch_brain.dataset.Dataset` ([docs](https://torch-brain--154.org.readthedocs.build/en/154/package/dataset/dataset.html#torch_brain.dataset.Dataset))

The `Dataset` class helps you load a generic dataset: A collection of temporaldata h5 files within a directory. So, basically , a single brainset.

```python
from torch_brain.dataset import Dataset

# Load all h5 files in a directory
ds = Dataset(dataset_dir="/path/to/processed/brainset")
ds
# >> Dataset(n_recordings=40)

# Load some recordings only
ds = Dataset(dataset_dir="/path/to/processed/brainset", recording_ids=["session1", "session2"])
ds
# >> Dataset(n_recordings=2)

# Explore recordings
ds.get_recording(ds.recording_ids[0])
# >> Data(...)
```

This class makes minimal assumptions about your data. For example, `ds.get_unit_ids()` does not exist, since we can't assume your data always has spiketrains - what if it was EEG, where the word "unit" is not canonical.

Each brainsets is somewhat unique, and the suggested way to handle this is by defining custom Dataset sub-classes for each brainset (although you could just work with base `Dataset` as well). And so, `torch_brain.dataset.Dataset` is designed to be easily extensible and overridable to encode properties and methods specific to your data. We also introduces [Mixins](https://stackoverflow.com/questions/533631/what-is-a-mixin-and-why-is-it-useful) that include common methods for different modalities.

To illustrate subclassing `Dataset`, let's make a simple spiking dataset, and also override the sampling intervals:
```python
from torch_brain.dataset import Dataset, SpikingDatasetMixin

class MySpikingDataset(SpikingDatasetMixin, Dataset):
    
    def __init__(self, root, dirname="my_spiking_dataset", recording_ids=None, transform=None):
        super().__init__(
            dataset_dir=Path(root) / dirname,
            recording_ids=recording_ids,
            transform=transform,
            namespace_attributes=["session.id", "subject.id", "units.id"],  # spiking dataset =>  we have units
        )
        
    def get_sampling_intervals(self, split: Literal["train", "valid", "test"]) -> dict[str, Interval]:
        if split == "train":
            # handle train splits
        elif split == "valid":
            # hand validation splits
        ...
        
    def get_recording_hook(self, data):
        # Override to do light changes to data objects returned by `MySpikingDataset.get_recording()`
        ...
        
ds = MySpikingDataset(root="/path/to/processed")

# SpikingDatasetMixin provides the `get_unit_ids()` and `compute_average_firing_rates()` methods
ds.get_unit_ids()
ds.compute_average_firing_rates()  # Just a thought. Open to removing if we don't agree about this mixin method
```

Mixins included:
- `SpikingDatasetMixin` provides the `get_unit_ids()` and `compute_average_firing_rates()` methods.
- `CalciumImagingDatasetMixin` is provides the `get_roi_ids()` method.

## Brainset-specific Dataset classes: `brainsets.datasets`

An accompanying PR in brainsets (neuro-galaxy/brainsets#67) adds these sub-classes customized for all existing brainsets. These classes will make it easier to explore existing brainsets. For example:

```python
from brainsets.datasets import PerichMillerPopulation2018

ds = PerichMillerPopulation2018(root="/path/to/processed")
ds.recording_ids
# >> ['c_20131003_center_out_reaching', 'c_20131009_random_target_reaching', ...]

# Since this dataset has a SpikingDatasetMixin, we can do
ds.get_unit_ids()  
# >> ['c_20131003_center_out_reaching/group_electrode_group_M1/elec0/unit_0', 
#    'c_20131003_center_out_reaching/group_electrode_group_M1/elec1/unit_1',
#    ...]

# Explore one session
ds.get_recording(ds.recording_ids[0])
# >> Data(...)
```

An additional benefit of doing this is to absorb the complexity of defining and resolving splits in a custom way for each dataset (another attempt at solving this: neuro-galaxy/brainsets#56).

For instance, the `FlintSlutzkyAccurate2012` dataset provides two different kinds of sampling interval splits:

```python
from brainsets.datasets import PerichMillerPopulation2018, FlintSlutzkyAccurate2012

ds = FlintSlutzkyAccurate2012(root="/path/to/processed", split_type=None)
ds.get_sampling_intervals("train")  # generic train split

ds = FlintSlutzkyAccurate2012(root="/path/to/processed", split_type="hand_velocity")
ds.get_sampling_intervals("train")  # split designed for hand velocity decoding task
```

Each brainset's Dataset is free to handle splits in its own way. These sub-classes are also pretty easy to define, have a look at neuro-galaxy/brainsets#67 to see the code!

**Note**: This idea of having separate classes for each dataset is coming from the way `torchvision` and `torch_geometric` handle their datasets. See [`torchvision.datasets`](https://docs.pytorch.org/vision/0.8/datasets.html) and [`torch_geometric.datasets`](https://pytorch-geometric.readthedocs.io/en/latest/modules/datasets.html).

## NestedDataset: `torch_brain.dataset.NestedDataset` ([docs](https://torch-brain--154.org.readthedocs.build/en/154/package/dataset/nested.html#torch_brain.dataset.NestedDataset))

One question remaining is: how will we represent Datasets that are made up of multiple brainsets? Using a `NestedDataset`!

```python
from torch_brain.dataset import NestedDataset
from brainsets.datasets import PerichMillerPopulation2018, FlintSlutzkyAccurate2012

ds_pm = PerichMillerPopulation2018(root="/path/to/processed")
ds_fs = FlintSlutzkyAccurate2012(root="/path/to/processed", split_type="hand_velocity")

# Combine in a NestedDataset
ds = NestedDataset([ds_pm, ds_fs])
ds.recording_ids
# >> ['FlintSlutzkyAccurate2012/monkey_c_e1_reaching', 'FlintSlutzkyAccurate2012/monkey_c_e2_reaching', ...
#     'PerichMillerPopulation2018/c_20131003_center_out_reaching', ...]

# 2nd way to create a NestedDataset
ds = NestedDataset({"perich_miller": ds_pm, "flint_slutzky": ds_fs})
ds.recording_ids
# >> ['flint_slutzky/monkey_c_e1_reaching', 'flint_slutzky/monkey_c_e2_reaching', ...
#     'perich_miller/c_20131003_center_out_reaching', ...]
```

And each `Dataset` and `NestedDataset` provides a point to add a transform. You can nest these datasets in whichever way you want. Want to nest the nested datasets? That's naturally supported.

image

---

### Potential future features:
- Hide `brainsets prepare` behind dataset instantiation. This will make our interface more similar to `torchvision`, `torch_geometric`, and `huggingface`. These packages are much easier to use because of this feature.
- Add "post_processing" or `pre_transform` support, similar to `torch_geometric`.
- Some folks have expressed interest in uploading/downloading splits to/from an S3 bucket to avoid RNG mishaps. This is already supported, and can be further streamlined.
- Some brainsets can add a `.validate()` method to check if the data looks okay.
- Add a proper schemas (dataclass) for ReadoutConfig and MultitaskReadoutConfig.
- Modality specific utility functions, such as `compute_average_firing_rates()` in `SpikingDatasetMixin`

### Concerns:

**What happens to old `Dataset`?**
The old dataset lives in `torch_brain.data` (as opposed to `torch_brain.dataset`, which is the proposed Dataset).  If this PR is approved, the old will be sunset in ~2 months, with DeprecationWarning helping users move to the new implementation. They both can co-exist for now.

**I like working with config files, do I _have_ to define my datasets in Python?**
No! Turns out, this implementation of Dataset can be translated well into a config that can be instantiated using `hydra.utils.instiate`. See [here](https://gist.github.com/vinamarora8/1b906add2aa305dfea300a6bff2f6f96) for an example implementation of POYO-1 dataset config with new `Dataset`. This is actually a better way to go about configs, because the Python classes naturally serve as a stable schema.

**Is the dataloading performance worse than before?**
No! I tested iterating over the POYO-MP and POYO-1 training datasets with no model in the loop. Here are runtimes to iterate over the dataset once with a random sampler ([script](https://gist.github.com/vinamarora8/7537f840dba73f5077a10d32685ef3c5)):
| dataset | Old `Dataset` | New `Dataset` |
| --------| -------------|--------------|
| POYO-1 train | 11m 51s  | 11m 18s      |
| POYO-MP train | 26s    | 22s |

----

#### TODO
- [x] Docstrings in `NestedDataset`
- [x] Include `NestedSpikingDataset` in docs
- [x] Add gist for example dataset config
- [x] Docstring for SpikingDatasetMixin
- [x] Add deprecation warning for old Dataset
- [x] Move POYO training to new Dataset
- [x] Performance comparison with old Dataset

## Summary by CodeRabbit

* **New Features**
  - Introduced a new dataset package with time-sliced HDF5 recording support, nested dataset composition, and mixin support for spiking and calcium-imaging workflows; added new example dataset flavors and explicit dataset presets.

* **Documentation**
  - Expanded Package Reference and added dedicated API pages and cross-link anchors for dataset and sampler docs.

* **Tests**
  - Added comprehensive unit tests covering dataset behaviors, nesting, mixins, and sampling.

* **Chores**
  - Updated example training configs, removed legacy dataset presets, and adjusted CI package installation source.

* **Deprecated**
  - Older dataset API now emits runtime deprecation warnings.


torch_brain/dataset/dataset.py
```diff
@@ -0,0 +1,238 @@
...
+        index = _ensure_index_has_namespace(index)
+
+        data = self.get_recording(index.recording_id, index._namespace)
+        sample = data.slice(index.start, index.end)
+        if self.transform is not None:
+            sample = self.transform(sample)
+        return sample
+
+    def get_sampling_intervals(self, *args, **kwargs) -> dict[str, Interval]:
+        """Get the time domain intervals for all recordings in the dataset.
```
milosobral: I fell like this docstring could be a bit clearer if it mentioned that these intervals are then used by the sampler.

torch_brain/dataset/dataset.py
```diff
@@ -0,0 +1,238 @@
...
+        """
+        index = _ensure_index_has_namespace(index)
+
+        data = self.get_recording(index.recording_id, index._namespace)
+        sample = data.slice(index.start, index.end)
+        if self.transform is not None:
+            sample = self.transform(sample)
+        return sample
+
+    def get_sampling_intervals(self, *args, **kwargs) -> dict[str, Interval]:
```
milosobral: What is the reason for taking *args and **kwargs here?

torch_brain/dataset/dataset.py
```diff
@@ -0,0 +1,238 @@
...
+        prefixed with ``ds1/``.
+
+        Subclasses can override :meth:`apply_namespace` to customize how a namespace is applied.
+    """
+
+    def __init__(
+        self,
+        dataset_dir: str,
+        recording_ids: Optional[list[str]] = None,
+        transform: Optional[TransformType] = None,
```
milosobral: I think a very common way of doing this will be to have alist of transforms instead of one single transform. Maybe we could also take care of that use case directly here by accepting Lists of TransformTypes here and putting them together to remove that responsibility from the user.

milosobral (commented):
Looked through it quickly but haven't tested anything yet

torch_brain/dataset/dataset.py
```diff
@@ -0,0 +1,238 @@
...
+        self._filepaths = {r: dataset_dir / f"{r}.h5" for r in self._recording_ids}
+        missing_files = [str(p) for p in self._filepaths.values() if not p.exists()]
+        if missing_files:
+            raise FileNotFoundError(
+                f"The following recording files do not exist: {missing_files}"
+            )
+
+        if keep_files_open:
+            self._data_objects = {
+                r: Data.from_hdf5(h5py.File(self._filepaths[r]))
```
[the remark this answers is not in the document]
vinamarora8: Yep, we could! It's not released yet tho :/

torch_brain/dataset/dataset.py
vinamarora8: I was thinking Dataset subclasses might want to have splits with more complex options than just a `Literal["train", "valid", "test"]`. I am open to fixing the signature to just the string literal, do you think we should?

torch_brain/utils/data.py
```diff
@@ -0,0 +1,31 @@
+from typing import Any
+from temporaldata import Data
+
+
+def set_nested_attribute_in_data(data: Data, path: str, value: Any) -> Data:
```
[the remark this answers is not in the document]
vinamarora8: Yes, it should be! But we'll have to make a release on temporaldata for it to propagate here. 
I have created an issue in temporaldata to track this neuro-galaxy/temporaldata#66.

torch_brain/transforms/container.py
```diff
@@ -19,9 +21,13 @@ def __init__(self, transforms: List[Callable]):
 
     def __call__(self, data: temporaldata.Data) -> temporaldata.Data:
         for transform in self.transforms:
-            data = transform(data)
+            if transform is not None:
```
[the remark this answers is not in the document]
vinamarora8: This is unrelated. My bad. I will remove it from this PR.

torch_brain/dataset/dataset.py
vinamarora8: done!

torch_brain/utils/misc.py
```diff
@@ -0,0 +1,18 @@
...
+    Adds a string prefix to each element of a numpy string array.
+
+    Args:
+        prefix (str): The string to prepend to each element.
+        array (np.ndarray): An array of strings or string-like objects.
+
+    Returns:
+        np.ndarray: New array with the prefix added to each element.
+    """
+    if np.__version__ >= "2.0":
```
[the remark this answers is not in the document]
vinamarora8: TODO!

docs/source/package/dataset/dataset.rst
```diff
@@ -0,0 +1,14 @@
+.. currentmodule:: torch_brain.dataset
```
milosobral: I think it would be worth it for these pages in the documentation to also have some extra explanation of the concepts and maybe even some examples. The API reference is super useful if you already understand what is going on and how things plug together, but I think having some extra explanation might make it easier for anyone to get started using the package.

milosobral (commented):
Honestly not many comments, it looks great!
I've been playing around with this branch for a few weeks now and haven't run into any issues so far so except for some of the nitpicky comments, I think we can think about merging that ASAP

torch_brain/data/dataset.py
```diff
@@ -92,6 +93,14 @@ def __init__(
         subject_id_prefix_fn: Callable[[Data], str] = default_subject_id_prefix_fn,
         keep_files_open: bool = True,
     ):
+
+        warnings.warn(
+            "torch_brain.data.dataset.Dataset is deprecated and will be removed in a future release. "
+            "Please switch to torch_brain.dataset.Dataset.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
```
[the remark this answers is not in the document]
vinamarora8: Good point. Done!

docs/source/package/dataset/dataset.rst
vinamarora8: Good point. I added some concept docs! These can definitely be expanded more in the future.  Lmk what you think!

torch_brain/dataset/dataset.py
```diff
@@ -0,0 +1,270 @@
...
+        if hasattr(self, "_data_objects"):
+            data = copy.deepcopy(self._data_objects[recording_id])
+        else:
+            fpath = self._filepaths[recording_id]
+            data = Data.from_hdf5(h5py.File(fpath))
+
+        self.get_recording_hook(data)
+        if _namespace:
+            self.apply_namespace(data, _namespace + "/")
+        return data
```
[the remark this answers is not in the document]
vinamarora8: Leaving file handle leakage concerns to be addressed in another PR.

milosobral (approved):
Looks good to me! Clear and concise explanation of the concepts I love it

PR #160, opened 2026-01-15
Improve numerical stability in bin_spikes with epsilon tolerance

AlexandreAndr:
### Summary
This PR improves the numerical robustness of spike binning by introducing a small tolerance parameter `eps` to the `bin_spikes` function. The change prevents floating-point precision issues when computing the number of bins for durations that are very close to an integer multiple of the bin size.

### Motivation
When `(end - start) / bin_size` is affected by floating-point roundoff, it can evaluate to a value slightly below an integer (e.g. 9.99999999 instead of 10.0). This may cause an extra bin to be dropped during truncation.
Adding a small epsilon before flooring stabilizes this computation and ensures consistent bin counts.

### Changes

- Added an `eps` parameter to `bin_spikes` to improve numerical stability when determining the number of bins.
- Updated docstrings and inline comments to document the motivation and behavior of `eps`.
- Added a regression test covering floating-point boundary conditions near exact bin multiples.
- Added corretion for spikes with domain not staring at 0 

### Backward Compatibility

- Default behavior is unchanged for well-behaved inputs.
- The new eps parameter has a default value and does not affect existing callers.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved numerical stability in spike binning so near‑boundary floating‑point timestamps no longer produce incorrect counts.
* **New Features**
  * Added a configurable stability option for binning to reduce rounding issues (defaults preserve prior behavior).
* **Tests**
  * Added an edge‑case test covering near‑boundary floating‑point timestamps to ensure consistent binning.
* **Documentation**
  * Updated changelog and docstrings to describe the new stability option and its impact.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,15 @@
 All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-- Added `eps` parameter to `bin_spikes` to improve numerical stability when computing bin boundaries.
+
+## [Unreleased]
+### Added
+- Added `eps` parameter to `bin_spikes` to improve numerical stability when computing bin boundaries. ([#160](https://github.com/neuro-galaxy/torch_brain/pull/160))
+
+### Removed
+
+### Changed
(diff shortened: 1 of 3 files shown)


vinamarora8 (approved):
Looks good!

PR #169, opened 2026-02-21
Refacto NDT2

AlexandreAndr:
Please review in priority on the modeling design `torch_brain/models/ndt2.py` and eventually `examples/ndt2/train.py` (not fully complete but gives and idea of the integration of the model)!  The others files will be updated soon!

TODO:

- [x] utils function to load the checkpoint (only the encoder in case of superv finetuning, ect)
- [x] match the initialization from original NDT2

Consolidates the NDT2 stack into a single configurable model with cleaner data/model boundaries and faster execution paths.

Key updates/feature:
- Unified architecture: one NDT2 class now handles both SSL pretraining and supervised decoding (replacing separate model implementations).
- Faster masking path: attention masks are built directly on GPU during forward pass to avoid extra host/device overhead.
- Externalized token masking: SSL keep/mask index sampling is moved to tokenization/collation, so the model receives explicit encoder/decoder token sets.
- Sample-level masking preserved: masking is still done per sample (not one shared batch mask), keeping training behavior aligned with prior setup.
- Configurable causality: supports both causal and non-causal attention mask construction via is_causal.
- token control: explicit context-token attention blocking to main-token stream where required.
- Cleaner supervised path: supervised mode reuses the same backbone and decoder flow with pooled latents + behavior query tokens.
- Improved maintainability: fewer codepaths, shared embeddings/transformers, and clearer forward/tokenize interfaces.

## Summary by CodeRabbit

* **New Features**
  * New NDT2 neural-data model supporting both self-supervised and supervised workflows with tokenized inputs.
  * End-to-end training workflow with configurable logging, checkpointing, LR scheduling, early stopping, multi-GPU support, per-session fragment training, and WandB integration.
  * New example configurations and dataset presets for IBL pretraining and supervised runs, including readout/normalization and evaluation metrics.
  * NDT2 exposed in the public model interface.


torch_brain/models/ndt2.py
```diff
@@ -0,0 +1,538 @@
...
+        ctx_tokens = []
+        if self.tokenize_session:
+            ctx_tokens.append(self.session_emb(session_idx) + self.session_bias)
+        if self.tokenize_subject:
+            ctx_tokens.append(self.subject_emb(subject_idx) + self.subject_bias)
+        if self.tokenize_task:
+            ctx_tokens.append(self.task_emb(task_idx) + self.task_bias)
+
+        if self.n_ctx_tokens > 0:
+            ctx_emb = torch.stack(ctx_tokens, dim=1)
```
vinamarora8: I don't think `self.n_ctx_tokens` is needed. `len(ctx_tokens)` should be equivalent

examples/ndt2/configs/train_ssl.yaml
```diff
@@ -0,0 +1,14 @@
...
+
+is_ssl: true
+wandb:
+  run_name: pretrain
+
+load_from_checkpoint: false
+new_decoder: false
+
+dataset:
+  _target_: datasets.mp.PoyoMPDataset
```
vinamarora8: @AlexandreAndr I think you forgot to commit the datasets directory in this PR

PR #171, opened 2026-02-26
Change from UnitFilterById to UnitFilterByAttr

AlexandreAndr:
Generalize unit filtering to support non-ID attributes (region/probe/etc.) while keeping ID as default

* **Motivation**
  * `UnitFilterById` is useful, but in practice we often need to filter units by attributes other than `units.id,` such as region, probe, or other metadata fields.

* **Details**
This PR makes the filtering API more flexible without changing the default behavior.
Introduced a more general attribute-based filter (`UnitFilterByAttr`) that can match on any unit attribute.
Kept default behavior aligned with current usage by defaulting to `filter_attr="id"`.
`UnitFilterById` is now conceptually a special case of the generalized API (`UnitFilterByAttr(..., filter_attr="id")`).

* **Tests Added**
 * `test_unit_filter_by_brain_region` verifies that filtering units by `brain_region == "M1"` keeps only matching units and correctly updates spike indices/timestamps when reindexing is enabled.
  * `test_unit_filter_reset_index_true_deterministic_remap` verifies that with `reset_index=True`, two separate sliced views produce the same deterministic unit-index remapping based on original unit order.
  * `test_unit_filter_reset_index_false_keeps_original_unit_index` verifies that with `reset_index=False`, spike events are filtered but their original unit indices are preserved and the units table is not reduced.

* **Documentation**
  * Added in the docs

## Summary by CodeRabbit

* **New Features**
  * Attribute-based unit filtering with keyword and regex patterns.
  * Flexible targeting: filter by one attribute while remapping/filtering another.
  * Deterministic remapping and improved alignment of unit indices and timestamps after filtering.

* **Documentation**
  * Updated docs to describe the new attribute-based filtering behavior.

* **Breaking Changes**
  * The unit-filtering API has been renamed; update any integrations accordingly.


torch_brain/transforms/unit_filter.py
```diff
@@ -1,85 +1,94 @@
...
 
 import numpy as np
-
 from temporaldata import ArrayDict, Data, IrregularTimeSeries, RegularTimeSeries
 
 
 class UnitFilter:
     r"""
-    Drop units based on the `mask_fn` given in the constructor.
+    Keep/drop units based on the `mask_fn` given in the constructor.
```
vinamarora8, line 11:
```suggestion
    Keep/drop units based on the ``mask_fn`` given in the constructor.
```

torch_brain/transforms/unit_filter.py
```diff
@@ -1,85 +1,94 @@
...
 
     Args:
-        mask_fn (Callable[[ArrayDict], np.ndarray]): A function that takes the unit ids and returns a mask to keep the units.
-        field (str): The field to apply the filter.
-        reset_index (bool, optional): If True, it will reset_index the unit index of the time series.
+        mask_fn (Callable[[Data], np.ndarray]): A function that takes the data object and returns a mask to keep the units.
+        target_attr (str): The attribute to apply the filter.
+        reset_index (bool, optional): If True, it will reset the unit index of the time series.
```
vinamarora8, line 16: Drop types

torch_brain/transforms/unit_filter.py
```diff
@@ -1,85 +1,94 @@
...
     """
 
     def __init__(
         self,
         mask_fn: Callable[[ArrayDict], np.ndarray],
```
vinamarora8, line 21:
```suggestion
        mask_fn: Callable[[Data], np.ndarray],
```

torch_brain/transforms/unit_filter.py
```diff
@@ -1,85 +1,94 @@
...
-        field: str,
+        target_attr: str,
```
vinamarora8, line 22:
```suggestion
        spikes_attr: str = "spikes",
        units_attr: str = "units",
```

I think either we let both be configurable, or we let neither be configurable

torch_brain/transforms/unit_filter.py
```diff
@@ -1,85 +1,94 @@
...
             if self.reset_index:
                 # hack to have the lookup array that remaps the unit index
                 relabel_map = np.zeros(original_num_units, dtype=int)
                 relabel_map[unit_mask] = np.arange(unit_mask.sum())
-                target_obj = getattr(data, self.field)
+                target_obj = data.get_nested_attribute(self.target_attr)
                 target_obj.unit_index = relabel_map[target_obj.unit_index]
 
         elif isinstance(target_obj, RegularTimeSeries):
             raise NotImplementedError("RegularTimeSeries is not supported yet.")
```
vinamarora8, line 52: Does "unit" dropout even make sense for RegularTimeSeries?

torch_brain/transforms/unit_filter.py
```diff
@@ -1,85 +1,94 @@
...
+        reset_index (bool, optional): If True, it will reset the unit index of the time series.
         keep_matches (bool, optional): If True, units matching the pattern will be kept.
     """
 
     def __init__(
         self,
+        target_attr: str,
         pattern: Union[str, Pattern],
-        field: str,
+        filter_attr: str = "id",
```
vinamarora8, line 79:
```suggestion
        filter_attr: str
```

I don't see why this should have a default value. `UnitFilterByAttr` sounds like user knows what attr they want to filter by

vinamarora8 (commented):
Better than FilterById 100%. Though honestly, I'm not really sure if we should keep having `"Unit*"` transforms. I believe we want to generalize to modalities. What do you think? Go with this, and think about more modalities later?

PR #173, opened 2026-03-03
Add MultiChannelDatasetMixin hook and export it

glchau:
Simple MultiChannelDatasetMixin for Neuroprobe 2025 being implemented here: https://github.com/neuro-galaxy/brainsets/pull/79 

## Summary by CodeRabbit

* **New Features**
  * Added sEEG dataset mixin with channel-ID retrieval and optional channel-ID uniquification across recordings.
* **Behavior**
  * Channel-ID handling supports configurable uniquification and validation; non-uniquified IDs remain disambiguated by recording.
* **Documentation**
  * API docs updated to include the new sEEG mixin.
* **Tests**
  * New tests covering sEEG scenarios, ID disambiguation, uniquification behavior, and error cases.

### DIFF
--- a/docs/source/package/dataset/mixins.rst
+++ b/docs/source/package/dataset/mixins.rst
@@ -17,11 +17,3 @@ Mixins
     :members:
     :show-inheritance:
     :undoc-members:
-
-.. autoclass:: SEEGDatasetMixin.ChannelView
-    :members:
-    :undoc-members:
-
-.. autoclass:: SEEGDatasetMixin.RecordingInfo
-    :members:
-    :undoc-members:
--- a/docs/source/package/dataset/mixins.rst
+++ b/docs/source/package/dataset/mixins.rst
@@ -17,7 +17,3 @@ Mixins
     :members:
     :show-inheritance:
     :undoc-members:
-
-Default channel-id uniquification behavior:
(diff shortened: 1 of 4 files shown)


vinamarora8 (commented):
Hey, thanks for the PR! Is there any other method that would be reasonable for general SEEG datasets? Something like `get_channel_ids` maybe?

If there is nothing general enough to add here, then I guess we don't need a Mixin for sEEG right now. What do you think?

glchau:
Thanks for the review! That's a good point. I am planning to implement more sEEG datasets soon, so it would be helpful to have the Mixin structure available for that. As for shared functions, I think we could do something along the lines of channel metadata and sampling rate. I'll add something there for that!

vinamarora8 (commented):
Leaving some comments on design.

Also, @milosobral would love for you to chip in. The idea for this Mixin is to be used for all *EEG modalities. Do you think it's a good idea to have it be common?

milosobral (commented):
Minor comment about docstring but otherwise I like it!

vinamarora8 (commented):
I think we are almost there! Leaving some comments

vinamarora8 (approved):
Looks good! Thanks for the edits!

PR #191, opened 2026-04-19
Vinam/pydata docs

vinamarora8:
- Sklearn inspired
- PyData sphinx theme
- Modular way to add sections in API reference
- Support module level docstrings
- Read `docs/source/api_reference.py` to understand what is going on


torch_brain/data/collate.py
```diff
@@ -351,3 +351,19 @@ def collate(batch):
...
+    "pad",
+    "pad8",
+    "pad2d",
+    "pad2d8",
+    "track_batch",
+    "track_mask",
+    "track_mask8",
+    "track_mask2d",
+    "track_mask2d8",
+]
```
divyansha1115, line 369: shouldn't `__all__`  be defined at top of the file?

torch_brain/models/__init__.py
```diff
@@ -2,5 +2,17 @@
 from .poyo_plus import POYOPlus
 from .capoyo import CaPOYO
 
-_classes = ["POYO", "POYOPlus", "CaPOYO"]
-_functions = ["poyo_mp"]
+__all__ = [
+    "POYO",
+    "POYOPlus",
+    "CaPOYO",
+]
```
divyansha1115, line 9: are we removing poyo_mp function from docs?

vinamarora8: I think having it as a function is weird. This breaks what torch_brain.models should be, in my head. 
It maybe should be a classmethod. What do u think?

torch_brain/data/collate.py
vinamarora8: It's okay to do it at the bottom. I have seen many examples of it being done both ways

torch_brain/models/__init__.py
divyansha1115: I agree we should remove this. But we can just remove the function in a separate PR. I don't think it's a good precedent to have the function but explicitly not have it in docs

I don't think poyo_mp should be there. It's a thing specific to the paper. No need for it to be in the library.

PR #196, opened 2026-04-21
Replace torch tensor operations with numpy in sampler iteration for 6-16x speedup

milosobral:
### Why

`RandomFixedWindowSampler`, `SequentialFixedWindowSampler`, and `DistributedStitchingFixedWindowSampler` were spending most of their time in per-element `tensor.item()` calls during window construction. That overhead becomes large at scale.

### What changed

- Replaced hot-path `torch.arange` + per-element `.item()` loops with numpy array generation and bulk concatenation in:
  - `RandomFixedWindowSampler.__iter__`
  - `SequentialFixedWindowSampler._indices`
  - `DistributedStitchingFixedWindowSampler._generate_indices`
- Kept sampler semantics intact (window placement rules, short-interval handling, and deterministic epoch-length logic).
- Added focused edge-case coverage in `tests/test_sampler.py` for random/sequential samplers.

### Why RNG bridging is important

`RandomFixedWindowSampler` accepts an optional `torch.Generator`. After moving random draws to numpy for performance, we bridge torch RNG state into numpy by sampling a seed from the provided `torch.Generator` and using `np.random.default_rng(seed)`. This matters because it preserves the contract that randomness is controlled by the sampler’s torch generator.

### Validation

- All sampler tests pass after changes (`tests/test_sampler.py`, `tests/test_stitcher_sampler.py`).
- Added new edge-case tests for interval boundaries, `drop_short`, large interval counts, and epoch-length consistency across seeds.
- Local benchmark showed substantial iteration speedups (roughly ~6x to ~16x depending on sampler and workload size).
- Initializing a `RandomFixedWindowSampler` to train on the entirety of `klinzing_sleep_ds005555` (about 1800 hrs of data) goes from taking around 40 seconds to about 2-5 seconds so huge speedup on large datasets with lots of sessions and intervals. (Note that this speedup is only felt at the beggining of traing, not once the epoch starts)

## Summary by CodeRabbit

* **Tests**
  * Added a reusable window-validation helper and broadened edge-case/property tests for all fixed-window samplers, including deterministic epoch checks and varied interval scenarios.
* **Refactor**
  * Internal sampling and window-generation logic rewritten for performance and robustness while preserving external behavior and interfaces.


torch_brain/data/sampler.py
```diff
@@ -79,63 +80,80 @@ def __len__(self):
...
+            if self.generator is not None
+            else None
+        )
+        rng = np.random.default_rng(rng_seed)
+
         for session_name, sampling_intervals in self.sampling_intervals.items():
+            sid = session_to_idx[session_name]
             for start, end in zip(sampling_intervals.start, sampling_intervals.end):
-                interval_length = end - start
+                interval_length = float(end - start)
```
vinamarora8: curious: why is this needed? is it to convert fp32 to fp64 or smth like that?

torch_brain/data/sampler.py
```diff
@@ -79,63 +80,80 @@ def __len__(self):
...
+        ends_chunks: list[np.ndarray] = []
+
+        # Bridge the torch Generator to a numpy RNG to avoid per-element
+        # .item() calls that dominate the cost of the original implementation.
+        rng_seed = (
+            int(torch.randint(0, 2**31 - 1, (1,), generator=self.generator).item())
+            if self.generator is not None
+            else None
+        )
+        rng = np.random.default_rng(rng_seed)
```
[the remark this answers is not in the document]
vinamarora8: @milosobral I'm thinking we should eventually change the API so that the user provides an integer seed instead of a torch generator. Not part of this PR though.

torch_brain/data/sampler.py
```diff
@@ -79,63 +80,80 @@ def __len__(self):
...
-                    right_offset = end - start - left_offset
+                    right_offset = interval_length - left_offset
 
-                # if there is one sample worth of data, add it
-                # this ensures that the number of samples is always consistent
-                if right_offset + left_offset >= self.window_length:
+                # Ensure epoch length stays consistent: if the left and right
+                # remainders together span a full window, emit one extra sample.
+                if right_offset + left_offset >= self.window_length - 1e-9:
                     if right_offset > left_offset:
```
vinamarora8, line 133: Why is the `- 1e-9` needed?

torch_brain/data/sampler.py
```diff
@@ -79,63 +80,80 @@ def __len__(self):
...
-                # sample a random offset
-                left_offset = (
-                    torch.rand(1, generator=self.generator).item() * self.window_length
+                left_offset = float(rng.random() * self.window_length)
+                starts = np.arange(
+                    float(start) + left_offset,
+                    float(end) - self.window_length + 1e-12,
+                    self.window_length,
+                    dtype=np.float64,
                 )
```
vinamarora8, line 119: Why is the `-1e-12` needed?

vinamarora8 (commented):
Great find! I had no idea torch RNG can underperform due to the need to call `.item()`.

Leaving some very early comments for just the RandomFixedWindowSampler

PR #203, opened 2026-05-12
cleanup!: remove `optim.py`

vinamarora8:
Since we have agreed to not serve custom optimizers with `torch_brain`, this PR removes `SparseLamb` and the `optim.py` module.


ShivashriganeshMahato (approved):
lgtm

PR #204, opened 2026-05-12
cleanup!: move FeedForward to model definition files

vinamarora8:
Moves `torch_brain.nn.FeedForward` to the models that need it. This `nn.Module` is opinionated and not flexible. It only offers GEGLU-based FFN which was used in POYO and POYO+.

FFNs of all different varieties exist and different variants are regularly used by ML researchers. In my experience, it is common for people to write their own FFNs. Thus, it would be best if we do not offer a singular FFN at all in our public interface.

If people want to use "the FFN that POYO used", they can still do:
```python
from torch_brain.models.poyo import FeedForward
```

Same goes for POYO+. I would suggest we follow this code pattern for all model-specific modules.


ShivashriganeshMahato (dismissed):
lgtm

ShivashriganeshMahato (approved):
lgtm^2

PR #205, opened 2026-05-12
cleanup: scripts/calculate_normalization_scales.py

vinamarora8:
Moved `scripts/calculate_normalization_scales.py` to examples/poyo_plus. The only place it is mentioned.


ShivashriganeshMahato (approved):
lgtm

PR #206, opened 2026-05-12
feat: Some simplifications in examples/poyo

vinamarora8:
We have moved the "official codebase for POYO" to [this link](https://github.com/nerdslab/poyo), and so can make the examples to more aimed at readability and understanding.

We can think of this as step-1 of simplification. It can be simplified even further, like using AdamW instead of SparseLamb (then we can remove optim.py). Also, we can just show one dataset and remove hydra configs.

**Main goals:**
- Small and readable training script (~150 lines of training code + ~100 lines of utils code)
- Pure PyTorch training loop (no Lightning)
- Free of `MODALITY_REGISTRY`

**Changes to `models/poyo.py`**:
- Replaced `readout_spec` constructor argument with a simple `dim_out` (that's all `readout_spec` was needed for).
- `.load_pretrained()` method updated to work with this ^
- Removed target-creation logic in the tokenizer. The tokenizer now only concerns itself with producing the input to the model. The `PoyoDatasetWrapper` takes care of producing the targets and query timestamps.
- New method: `POYO.init_vocabs` which takes in a `Dataset` and creates the embedding table vocabs.
- Removed `poyo_mp()` function because it was stupid.

**New things introduced:**
These things are only introduced within `examples/poyo` and not as torch_brain API. They could be torch_brain API _some day_.
- `examples.poyo.utils.BehaviorStitcher`: A simple class to do what `DecodingStitchingEvaluator` was doing. This implementation has a "one-stitcher-per-session" philosophy. See the validation epoch in the train script to understand further.
- `examples.poyo.dataset.PoyoDatasetWrapper` takes care of producing the targets and query timestamps.

⚠️ **WARNING**: The new simple training script does not support multi-GPU training. I think this is OK (?) since the job of examples should be to just show users how to user our core features.

**Tests:**
Training curves seem to match well up to RNG (red=lightning, purple=this PR):
image


examples/poyo/utils.py
```diff
@@ -0,0 +1,110 @@
...
+        f"# Non-Embedding params={sum(p.numel()for p in nonemb_params):,}"
+    )
+    return optim, scheduler
+
+
+def weighted_mse_loss_fn(pred, target, weights):
+    """Simple sample-weighted MSE loss"""
+    loss = torch.nn.functional.mse_loss(pred, target, reduction="none")
+    loss = loss.flatten(1).mean(1) * weights
+    loss = loss.sum() / weights.sum()
```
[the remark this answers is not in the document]
vinamarora8: we'll keep the functionality of the old code.

divyansha1115 (commented):
I think the code is still fairly complex for someone who is new to the field.
I’d suggest removing Hydra (if feasible), behavior stitchers, and other components that are primarily there to improve POYO performance.

In my view minimal code would just be POYO model, with a very standard setup that will be applicable to any supervised model.

vinamarora8:
@divyansha1115 I agree that it can be simplified a lot more. I wanted to do this in two steps, as mentioned in the PR description. 

This PR primarily removes dependency on `MODALITY_REGISTRY`, `nn/loss.py`, and Lightning. I will open another PR soon on top of this to take the simplification all the way

divyansha1115 (dismissed):

CHANGELOG.md
```diff
@@ -22,6 +22,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Replaced `torch_brain.data.dataset.DatasetIndex` with `torch_brain.dataset.DatasetIndex` ([#186](https://github.com/neuro-galaxy/torch_brain/pull/186))
 - `bin_spikes` output shape changed from `(N, T)` to `(T, N)` and default `dtype` changed from `np.float32` to `np.int32` ([#170](https://github.com/neuro-galaxy/torch_brain/pull/170))
 - Renamed `torch_brain.models.CaPOYO` to `torch_brain.models.CalciumPOYOPlus` ([#198](https://github.com/neuro-galaxy/torch_brain/pull/198))
+- POYO API changes ([#206](https://github.com/neuro-galaxy/torch_brain/pull/206)):
```
divyansha1115, line 26: Add the removal of poyo_mp to changelog

torch_brain/models/poyo.py
```diff
@@ -9,6 +10,7 @@
 from torchtyping import TensorType
 from temporaldata import Data
 
+from torch_brain.dataset import Dataset
 from torch_brain.data import chain, pad8, track_mask8
```
divyansha1115: chain is a dead import. remove it?

torch_brain/models/poyo.py
```diff
@@ -67,7 +66,7 @@ def __init__(
         self,
         *,
         sequence_length: float,
-        readout_spec: ModalitySpec,
+        dim_out: int,
```
divyansha1115: dim_out is not present in the doc string of model

vinamarora8: great catch!

divyansha1115 (dismissed):
lgtm

vinamarora8:
Thanks for the review @divyansha1115. I had to push another commit to fix the docstring of POYO not rendering correctly.

divyansha1115 (dismissed):

divyansha1115 (approved):
lgtm!

PR #211, opened 2026-05-14
chore: remove hydra-core from core dependencies

### LINKED ISSUE #210: Unneeded dependency: `hydra-core`
The dependency `hydra-core` is listed as a core dependency in pyproject.toml, but it is only needed in our two examples. Thus, this can be removed.

vinamarora8:
`hydra-core` has only needed been for examples, and so can be removed from our core dependencies.

Additionally, `omegaconf` was needed for only the deprecated `torch_brain.data.Dataset`, and so I am adding it to the dev dependencies. Plus, an error to guide users if they are still using old `Dataset` and do not have `omegaconf` installed.

OmegaConf is also used in one of our docs page. That docs page needs to be updated. `omegaconf` in docs dependencies will handle it for now.

Closes #210

### DIFF
--- a/torch_brain/data/dataset.py
+++ b/torch_brain/data/dataset.py
@@ -112,7 +112,7 @@ class Dataset(torch.utils.data.Dataset):
         except:
             raise ImportError(
                 "Could not find `omegaconf` which is needed for the old Dataset."
-                " Please install using `pip intall omegaconf`."
+                " Please install using `pip install omegaconf`."
                 " Alternately, it is recommended to switch to torch_brain.dataset.Dataset."
             )
 
--- a/torch_brain/data/dataset.py
+++ b/torch_brain/data/dataset.py
@@ -109,7 +109,7 @@ class Dataset(torch.utils.data.Dataset):
 
         try:
             import omegaconf
-        except:
+        except ImportError:
             raise ImportError(
                 "Could not find `omegaconf` which is needed for the old Dataset."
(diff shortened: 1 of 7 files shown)


torch_brain/data/dataset.py
```diff
@@ -108,6 +107,15 @@ def __init__(
         self.subject_id_prefix_fn = subject_id_prefix_fn
         self.keep_files_open = keep_files_open
 
+        try:
+            import omegaconf
+        except:
+            raise ImportError(
+                "Could not find `omegaconf` which is needed for the old Dataset."
+                " Please install using `pip intall omegaconf`."
```
divyansha1115: typo intall -> install

divyansha1115 (dismissed):

torch_brain/data/dataset.py
```diff
@@ -108,6 +107,15 @@ def __init__(
         self.subject_id_prefix_fn = subject_id_prefix_fn
         self.keep_files_open = keep_files_open
 
+        try:
+            import omegaconf
+        except:
```
divyansha1115: it would be better to make this except ImportError.

vinamarora8: I am raising an ImportError here right?

vinamarora8: Ok just understood what you were saying. Fixed!

divyansha1115 (approved):
lgtm

PR #214, opened 2026-05-16
docs: Minimal training example on NLB Maze brainset

vinamarora8:
Adds `examples/nlb_maze_minimal` and a corresponding `docs/guides/examples` [page](https://torch-brain--214.org.readthedocs.build/en/214/generated/notebooks/nlb_maze_minimal_example.html)

**Goal:** Present an actually minimal training example using torch_brain. The train script is ~150 lines of code, and additional ~100 line file, models.py, contains 3 model definitions. The docs include a notebook that also presents some simple visualizations of the model's I/O.

**Notes:**
- The "Open in Colab" links would work after this PR gets merged (they rely on web paths that will only exist after merge)
- The notebook modification script is needed to hide some cells, and add the "Open to Colab" links

-----------

**Previous description (IGNORE THIS):**

Adds `examples/nlb_maze_minimal`,  a Google collab [notebook](https://colab.research.google.com/drive/1r1vbxmqccgHz-6det9Bxk3Ld6xU8RZdk?usp=sharing), and a corresponding `docs/guides/examples` [page](https://torch-brain--214.org.readthedocs.build/en/214/generated/nlb_maze_train.html)

**Goal:** Present an actually minimal training example using torch_brain. The train script is ~150 lines of code, and additional ~100 line file, models.py, contains 3 model definitions. The docs include a notebook that also presents some simple visualizations of the model's I/O.

The notebook-on-docs works like this:
- Docs build process downloads the notebook from google collab
- It uses myst_nb to display the notebook as a doc page


docs/source/nlb_maze_notebook.py
```diff
@@ -0,0 +1,71 @@
...
+    a warning is emitted and the download is skipped so the docs build can
+    still complete offline."""
+    import json
+    import socket
+    import urllib.error
+    import urllib.request
+
+    import nbformat
+
+    colab_id = "1r1vbxmqccgHz-6det9Bxk3Ld6xU8RZdk"
```
divyansha1115: is the colab in your personal drive @vinamarora8? if yes maybe we should have a separate google account for torch brain related tutorials

examples/nlb_maze_minimal/train.py
```diff
@@ -0,0 +1,160 @@
...
+        # training and validation trials. select_by_mask is the standard way
+        # to filter an Interval down to a subset based on a boolean mask.
+        trial_split_indicator = recording.trials.split_indicator.astype(str)
+        train_trials = trials.select_by_mask(trial_split_indicator == "train")
+        val_trials = trials.select_by_mask(trial_split_indicator == "val")
+
+        if self.split == "train":
+            return {rid: train_trials}
+        elif self.split == "val":
+            return {rid: val_trials}
```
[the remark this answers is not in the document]
divyansha1115: I think you should add this. Just in case someone is wondering why we don't use test.

examples/nlb_maze_minimal/train.py
```diff
@@ -0,0 +1,159 @@
...
+        # In this example, we have designed all models in `model.py` to take in
+        # a Tensor of shape (Number of neurons, Bins), and return a Tensor of shape
+        # (Number of output timestep, Output dimension).
+
+        # Spikes are an irregular event stream — bin them into a regular grid.
+        X = bin_spikes(data.spikes, num_units=len(data.units), bin_size=self.bin_size)
+        X = torch.from_numpy(X).float()  # shape: (num_bins, num_units)
+
+        # Hand velocity is already a regularly-sampled signal, so we just rescale.
+        Y = data.hand.vel / 200.0  # approximate z-score normalization
```
divyansha1115, line 96: is this really near what z-score normalization would be? if not it might be better to just leave it at saying we rescale so the loss is in a good range.

divyansha1115 (dismissed):
Looks good to me! I left some minor comments.

examples/nlb_maze_minimal/train.py
vinamarora8: It is near zscore values

docs/source/nlb_maze_notebook.py
vinamarora8: I don't think separate google account is a good idea. Google has it made it very hard to share a single google account with all the 2FA checks. I can move this into a shared drive. Although, that is not needed either, because if someone wants to update this, they could copy the notebook into their accnt and make the changes + update the link in this repo.

examples/nlb_maze_minimal/train.py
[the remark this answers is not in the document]
vinamarora8: umm.. raising errors in code that is meant for explanation does not seem right to me

docs/source/nlb_maze_notebook.py
vinamarora8: Moving to a shared drive makes more sense after thinking a bit. It would be so that I don't accidentally delete/modify the source file. 
We can figure out how to properly version control this stuff as we go forward!

examples/nlb_maze_minimal/train.py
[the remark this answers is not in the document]
divyansha1115: can u add a comment about why we dont have test. It looks weird to not have test in a minimal example.

docs/source/nlb_maze_notebook.py
divyansha1115: sure shared drive makes sense!

examples/nlb_maze_minimal/train.py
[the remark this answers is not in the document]
vinamarora8: What could a reason be? :3

vinamarora8:
We have decided to upload the notebook itself to the docs directory.
Plan is to
1. Use [ReviewNB](https://www.reviewnb.com/) to review future notebook diffs
2. Keep the "Open in Colab" button using this [Colab feature](https://colab.research.google.com/github/googlecolab/colabtools/blob/master/notebooks/colab-github-demo.ipynb)

vinamarora8:
@divyansha1115 Thanks for the previous review. I have added a comment + assert statement for why we are only showing train and val (and not test). 

Plus, I have made this notebook local + added some tooling for it to be shown well on the docs. See the updated PR description

divyansha1115 (approved):
looks great!

PR #217, opened 2026-05-19
Move `torch_brain.data.sampler` to `torch_brain.samplers`

AlexandreAndr:
## Summary

- Extracts sampler classes from `torch_brain.data.sampler` into a new dedicated `torch_brain.samplers` module, with each class in its own file
- Adds complete API documentation to all samplers (RST cross-references, `Example::` blocks, `Args`/`Returns`/`Raises` sections) following the `torch_brain.dataset` style
- **Bug fix:** `DistributedEvaluationSamplerWrapper.rank_len()` had an inverted comparison (`(total % num_replicas) < rank` → `rank < (total % num_replicas)`), causing `len()` to report the wrong count for any rank when the total sample count is not evenly divisible by `num_replicas`

## New module structure

| Class | Description |
|---|---|
| `RandomFixedWindowSampler` | Random fixed-length windows with per-epoch jitter |
| `SequentialFixedWindowSampler` | Deterministic fixed-length windows for evaluation |
| `TrialSampler` | One sample per pre-defined trial interval |
| `DistributedEvaluationSamplerWrapper` | Wraps any sampler for distributed eval without dropping samples |
| `DistributedStitchingFixedWindowSampler` | Co-locates overlapping windows on the same rank for local stitching |

`torch_brain.data.sampler` is kept as a thin re-export shim for backwards compatibility.


torch_brain/data/sampler.py
```diff
@@ -1,505 +1,16 @@
...
 }
+
+
+def __getattr__(name):
+    if name in _MOVED:
+        raise ImportError(
+            f"`torch_brain.data.sampler.{name}` has moved to `torch_brain.sampler`. "
+            f"Use `from torch_brain.sampler import {name}` instead."
+        )
+    raise AttributeError(f"module 'torch_brain.data.sampler' has no attribute {name!r}")
```
vinamarora8, line 16: I think you can just raise an import error on any `__getattr__`, don't have to check if it is a valid sampler. 
```suggestion
def __getattr__(name):
    raise ImportError(
        f"`torch_brain.data.sampler` module has moved to `torch_brain.sampler`. "
        f"Use `from torch_brain.sampler import {name}` instead."
    )
```

torch_brain/sampler/random_fixed_window.py
```diff
@@ -0,0 +1,170 @@
...
+        self.generator = generator
+        self.drop_short = drop_short
+
+    @cached_property
+    def _estimated_len(self) -> int:
+        num_samples = 0
+        total_short_dropped = 0.0
+
+        for intervals in self.sampling_intervals.values():
+            for start, end in intervals:
```
vinamarora8: Can we keep the `for start, end in zip(intervals.start, intervals.end)` from before?
I think the (start, end) pair iterator will have to be removed if we want to start supporting `Interval.__getitem__()` (like we talked offline yesterday). A for-loop like iterator this is inconsistent with the indexing we were thinking

torch_brain/sampler/sequential_fixed_window.py
```diff
@@ -0,0 +1,128 @@
...
+        >>> sampling_intervals = {
+        ...     "session_1": Interval(
+        ...         start=np.array([0.0]),
+        ...         end=np.array([100.0]),
+        ...     ),
+        ...     "session_2": Interval(
+        ...         start=np.array([0.0]),
+        ...         end=np.array([100.0]),
+        ...     ),
+        ... }
```
vinamarora8: Can use the shorter syntax
```suggestion
        ...     "session_1": Interval(0., 100.),
        ...     "session_2": Interval(0., 100.),
        ... }
```

torch_brain/__init__.py
```diff
@@ -5,6 +5,7 @@
 from . import utils
 from . import transforms
 from . import registry
+from . import sampler
```
vinamarora8, line 9: I think we should call this module `torch_brain.samplers`

vinamarora8 (commented):
Looking good! I think we should call this module `torch_brain.samplers` (as in, _plural_). Plus, left a few more comments ^

torch_brain/data/sampler.py
```diff
@@ -1,505 +1,5 @@
...
-# see docs/source/api_reference.py
-__api_ref__ = {
-    "description": "See :ref:`sampling` for further details.",
-    "sections": [{"autosummary": __all__}],
-}
+def __getattr__(name):
+    raise ImportError(
+        f"`torch_brain.data.sampler.{name}` has moved to `torch_brain.samplers`. "
+        f"Use `from torch_brain.samplers import {name}` instead."
+    )
```
vinamarora8:
```suggestion
def __getattr__(name):
    raise ImportError(
        f"`torch_brain.data.sampler` has moved to `torch_brain.samplers`. "
        f"Use `from torch_brain.samplers import {name}` instead."
    )
```

Since user could try to import something that was never in torch_brain.data.sampler in the first place

vinamarora8: I think docs pytest is now failing because I asked you to remove `_MOVED`. Sorry about that! It'll have to put back again

vinamarora8 (commented):
Looks good! One minor issue

vinamarora8:
@AlexandreAndr, also the `[tool.setuptools]` section of `pyproject.toml` will need to be updated with the new module. (we should change this in the future so that setuptools auto-exports all modules)

vinamarora8 (approved):
looks good!

PR #222, opened 2026-05-24
chore: rm notebooks/

vinamarora8:
The current `notebooks/model_writing_tutorial.ipynb` is quite old now, and relies on the old `Dataset`.
The new NLB Maze Minimal example would serve as a better starting point for people. It does not talk about "tokenizers," and that is something we should add with another minimal example.


AlexandreAndr (approved):
LGTM

PR #226, opened 2026-05-25
Remove Loss Module

AlexandreAndr:
# Summary

- Removed `torch_brain.nn.loss` module (`Loss`, `MSELoss`, `CrossEntropyLoss`, `MallowDistanceLoss`)
- Deleted `tests/test_loss.py`
- Registry now uses `torch.nn.MSELoss(reduction="none")` and `torch.nn.CrossEntropyLoss(reduction="none")` directly
- Weighted reduction logic moved into the POYO+ training loop

## Note

The changes to `registry.py` and `examples/poyo_plus/train.py` are the minimal required to keep the existing training pipeline working without the removed module. The registry is expected to be removed too.


vinamarora8 (commented):
I think `registry.py` needs to go before `nn.loss`? Or both at the same time.
Doing it loss-first leaves `registry.py` in limbo, where the loss fns are edited and are not the same as before (non-weighted). Might as well remove `registry.py` at that point

PR #227, opened 2026-05-26
cleanup!: rm seed_everything

AlexandreAndr:
`torch_brain.utils.seed_everything` moved out of the public API; a reference implementation is still available in `examples/poyo/utils.py` and `examples/poyo_plus/utils.py`. 
Calling it now raises an `ImportError` pointing to the POYO example.


vinamarora8 (approved):
LGTM!

PR #231, opened 2026-05-27
cleanup!: Remove torchtyping

AlexandreAndr:
## Summary

Removes `torchtyping` as a required dependency. The library's runtime
tensor-shape assertions were not being used (no `@typechecked` decorator),
so the annotations provided no functional benefit over plain `torch.Tensor`.

- Remove `torchtyping~=0.1` from `pyproject.toml` dependencies
- Replace all `TensorType[...]` annotations in `POYO.forward` with
  `torch.Tensor` / `torch.Tensor | None` using modern union syntax
- Clean up docstrings to use concise `(B, N)` shape notation
- Add explicit type hints to `FeedForward.__init__` parameters
- Drop now-unused `Optional` and `Union` imports

### DIFF
--- a/torch_brain/models/poyo.py
+++ b/torch_brain/models/poyo.py
@@ -1,16 +1,16 @@
 import inspect
-from typing import Dict, List
 import logging
-
 from pathlib import Path
+from typing import Dict, List
+
 import numpy as np
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
 from temporaldata import Data
 
-from torch_brain.dataset import Dataset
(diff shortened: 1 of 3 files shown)


torch_brain/models/poyo.py
vinamarora8, line 1: Can use :math: in the docstring shape pls

AlexandreAndr: Good catch!

vinamarora8 (approved):
LGTM!!

PR #232, opened 2026-05-27
refactor: move collate utilities to torch_brain.batching

AlexandreAndr:
## Summary

- Introduces a new `torch_brain.batching` module as the canonical home for all collate utilities (`collate`, `chain`, `pad`, `pad8`, `pad2d`, `pad2d8`, `track_batch`, `track_mask`, `track_mask8`, `track_mask2d`, `track_mask2d8`).
- Replaces `torch_brain.data.collate` with a stub that raises a descriptive `ImportError` pointing users to the new location — no silent breakage.
- Removes the flat re-exports of collate symbols from `torch_brain.data.__init__`; `torch_brain.data` now only exports `Dataset`.
- Updates examples, tests, and API reference docs accordingly.

## Migration

```python
# Before
from torch_brain.data import collate, pad, chain, ...
from torch_brain.data.collate import collate, pad, chain, ...

# After
from torch_brain.batching import collate, pad, chain, ...

### CODE REVIEW

File: docs/source/api_reference.py [hunk 1]
```diff
@@ -57,6 +51,7 @@
 
 # Modules to include in API reference.
 API_MODS = [
+    "torch_brain.batching",
     "torch_brain.dataset",
```
vinamarora8: batching as the first module is a bit weird. Move this below samplers? And move torch_brain.data to the end

vinamarora8 (commented):
Minor comment. Looks good otherwise

docs/source/api_reference.py
AlexandreAndr: Yes you are right I did not tought about the impact on the docs I'll update it!

AlexandreAndr:
An additional wuestion I add was about a potential file splitting. Would we like to split them as we do with sampler or should we keep it like that? Maybe we can keep like that for now and reorganize if the files grows more?

vinamarora8:
> An additional wuestion I add was about a potential file splitting. Would we like to split them as we do with sampler or should we keep it like that? Maybe we can keep like that for now and reorganize if the files grows more? 

Yes we should eventually split and organize these functions. But it is a separate cleanup task and can be done later!

vinamarora8 (approved):
LGTM alex!

PR #233, opened 2026-05-30
Add GOVERNANCE.md

vinamarora8:


AlexandreAndr:
I am all good with the governance charter! It might evolve in time as the project expand (e.g., creation of new roles) but it is good to have a starting point! 
I'll approve the PR but can we also wait for @milosobral appoval before we merge the PR?

AlexandreAndr (approved):

vinamarora8:
@AlexandreAndr Milo was good with the charter before he went on vacation, and it has not changed since then. So, if I'm not wrong, everyone involved has had a look at this exact version and OK'd it.

PR #251, opened 2026-06-01
Cleaning temporaldata and brainsets references or imports

AlexandreAndr:


vinamarora8 (commented):
Looks mostly great! Amazing work! All CI green is amazing to see at this stage

There are some issues in docs/, but those will be overwritten by a general update of the User Guide (#253), so not worth addressing here.

vinamarora8 (approved):

vinamarora8:
This also closes #248 right?

PR #254, opened 2026-06-02
docs: Improve user-guide for `torch_brain.data`

### LINKED ISSUE #253: Docs: New flow for `torch_brain.data` user guide
Partially addressed in #254. The text needs some some rewording. Follow Milo's [suggestion](https://github.com/neuro-galaxy/torch_brain/pull/254#pullrequestreview-4428485059)

#281 should be the final chapter in this issue

vinamarora8:
Read the docs here: https://torch-brain--254.org.readthedocs.build/en/254/
This PR mainly edits the "[The Data Format](https://torch-brain--254.org.readthedocs.build/en/254/guides/data/index.html)" section of the User Guide

Side edits:
- Organized the `docs/source/guides` directory
- Moved around some doc pages for flow after data user guide
- Set default domain or IrregularTimeSeries to "auto" (hard to explain wtf domain is in the first user guide)

This PR is focused on addressing the user guide for the `torch_brain.data` module. Previous docs for this section were somewhat antiquated: sometimes because of `temporaldata` -> `torch_brain.data` change, other times due to some new features that have simplified the API from when the docs were originally written. Plus, I have tried to focus more on teaching the reader how to use our stuff, and have skipped talking about some advanced stuff entirely (e.g. the `from_dataframe()` method).

> [!NOTE]
> I am not super happy about the "Domains" user guide. But we can improve it later?

Closes #253


docs/source/guides/data/interval_ops/index.rst
```diff
@@ -0,0 +1,250 @@
...
+
+
+Dilation
+--------
+
+The :obj:`~Interval.dilate` method expands each interval by a specified amount
+on both sides.
+
+.. code-block:: pycon
+
```
AlexandreAndr, line 82: Add
```
>>> # Create three intervals [1., 5.), [10., 13.5), and [14., 18.)
```

docs/source/guides/data/interval_ops/index.rst
```diff
@@ -0,0 +1,250 @@
...
+We also provide some introspection methods to check whether the periods in an
+:obj:`Interval` are *disjoint* (non-overlapping) and *sorted* (in increasing
+order of start time).
+
+Here, the two periods ``[0, 1.1)`` and ``[1, 2)`` overlap, so the interval is
+not disjoint, but its start times are still in increasing order:
+
+.. code-block:: pycon
+
+   >>> interval = Interval(start=[0., 1.], end=[1.1, 2.0])
```
AlexandreAndr, line 152: Add
```
# Create two intervals [1., 1.1), and [1., 2.)
```

docs/source/guides/data/interval_ops/index.rst
```diff
@@ -0,0 +1,250 @@
...
+   >>> interval.is_disjoint(), interval.is_sorted()
+   (False, True)
+
+By contrast, these periods don't overlap and are already ordered, so both
+checks return ``True``:
+
+.. code-block:: pycon
+
+   >>> interval = Interval(start=[0., 3.], end=[1., 4.])
```
AlexandreAndr, line 162: Add
```
>>> # Create two intervals [0., 1.), and [3., 4.)
```

docs/source/guides/data/interval_ops/index.rst
```diff
@@ -0,0 +1,250 @@
...
+   Visualization of the coalesce operation
+
+
+The coalesce operation is useful for:
+
+- Cleaning up noisy interval data
+- Merging intervals that are effectively continuous
+- Simplifying interval representations
+
+There are multiple edge cases that can occur when performing interval
```
AlexandreAndr: Do we want to wrap this into a warining section?

AlexandreAndr:
Looks great to me just added minor comments!

> I am not super happy about the "Domains" user guide. But we can improve it later?

I agree we can imporve later. I think for the domain in general, we probably need to rework on it for irregular_ts with the right edge on `"auto"` domain (last timestamps excluded from domain in theory with the `[start, end)` convention)

docs/source/guides/data/interval_ops/index.rst
vinamarora8: I put it under a note. Warning felt a bit much for an introduction doc :3

divyansha1115 (commented):
would be nice to add a reference to gappy when talking about regular ts in the first doc.

vinamarora8:
I have refactored the "Meet the Data Objects" page. 

- ArrayDict is presented last (as it seems like a boring way to introduce the data module). Also @divyansha1115 suggested.
- All objects have admonitions suggesting alternate ways to construct the objects.
- Add a "Conventions" sub-section

divyansha1115 (approved):
lgtm

milosobral (commented):
I like the organization and flow of that and I think the explanations themselves are great. However: 
- I'm personally not a big fan of the "tutorial/conversational" tone for some of the explanations. I feel like most times when I read a guide, I want the shortest path possible to me getting enough information to start using the package. I think it would be better to simplify some of the wording to avoid things like: 'As the name suggests' or 'In other words'. 
- I also think sometimes the different tenses get a bit mixed up. I think we should unify everything to use the active present tense. Here is an example of what I mean: "In this example, slicing spikes from 2 to 4 seconds should give us the 2 spikes that fall within that window." --> "We slice the data from 2 to 4 seconds which returns the 2 spikes within that window."

I think the PyTorch guides do a great job of having some level of flow while using a very straightforward declarative style of writing. Here is a random example from a random sentence from a random guide: "We define our neural network by subclassing nn.Module, and initialize the neural network layers in __init__. Every nn.Module subclass implements the operations on input data in the forward method."

vinamarora8:
@milosobral I really like your suggestion! torch's guides are the standard that I would also like to follow. 

I will modify these later in the evening today. I do not want to further block your work on `torch_brain.pipeline` docs though. So we could merge this right now and I rephrase things tonight/tomorrow. Sounds ok?

milosobral (approved):

PR #259, opened 2026-06-02
rm `np.array` int the data object initialization in the docs

AlexandreAndr:


vinamarora8:
@AlexandreAndr I'm working on #253 (and PR #254, currently draft), which will solve this

AlexandreAndr:
Perfect!

PR #262, opened 2026-06-02
merge: Modernize typing and replace black with ruff

### LINKED ISSUE #260: Update some typing (e.g., `Dict` -> `dict`...)

### LINKED ISSUE #261: Add ruff as a new linter

AlexandreAndr:
## Summary
- Replaced all legacy `typing` imports (`Optional`, `Dict`, `List`, `Tuple`, `Union`) with modern Python 3.10+ syntax (`X | None`, `dict`, `list`, etc.) using `pyupgrade`
- Replaced `black` with `ruff` for formatting and linting — covers import sorting (`I`), unused imports (`F`), pyupgrade rules (`UP`), bugbear (`B`), and pycodestyle (`E`); updated pre-commit hook and CI accordingly
- Cleaned up docstrings: removed inline `(type)` annotations from `Args:` sections and `np.array([...])` wrappers from examples
- Fixed circular imports in `torch_brain/datasets/` by switching to relative imports (`.dataset`, `.mixins`) instead of going through the package `__init__.py`
- Fixed ~200 ruff violations across `torch_brain/`, `tests/`, `brainsets_pipelines/`, `docs/` (unused imports, missing `raise ... from err`, missing `stacklevel` on `warnings.warn`, unused loop variables)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,4 +3,5 @@ repos:
     rev: v0.15.15
     hooks:
     - id: ruff
+      args: [--fix]
     - id: ruff-format
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
-  - repo: https://github.com/psf/black
-    rev: 24.3.0
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.15.15
     hooks:
-    - id: black
-      language_version: python3
+    - id: ruff
+    - id: ruff-format
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,6 @@ dev = [
     "torch~=2.0",
     "scipy>=1.10.1",
     "pytest<9",
-    "black==24.2.0",
     "pre-commit>=3.5.0",
     "omegaconf",  # TODO: only needed for old Dataset, remove after cleanup
     "flake8",
@@ -48,6 +47,7 @@ dev = [
     "mne-bids",
     "dandi",
     "ty",
+    "ruff",
 ]
 docs = [
     "sphinx",
@@ -75,6 +75,12 @@ include = ["torch_brain*"]
 
 [project.scripts]
 brainsets = "torch_brain.pipeline._cli:cli"
+
+[tool.ruff]
+target-version = "py310"
+
+[tool.ruff.lint]
+select = ["I"]
 
 [tool.ty.environment]
 python-version = "3.10"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -78,7 +78,7 @@ brainsets = "torch_brain.pipeline._cli:cli"
 
 [tool.ruff]
 target-version = "py310"
-exclude = ["brainsets", "temporaldata"]
+exclude = ["brainsets", "temporaldata"] # TODO: remove after cleanup
 
 [tool.ruff.lint]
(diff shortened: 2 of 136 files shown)


AlexandreAndr:
The only things that might not look great is some string jump lines

pyproject.toml
```diff
@@ -76,6 +76,14 @@ include = ["torch_brain*"]
 [project.scripts]
 brainsets = "torch_brain.pipeline._cli:cli"
 
+[tool.ruff]
+target-version = "py310"
+exclude = ["brainsets", "temporaldata"] # TODO: remove after cleanup
+
+[tool.ruff.lint]
+select = ["I", "F", "UP", "B", "E"]
+ignore = ["E501", "B905"]
```
milosobral, line 85: Do we need to ignore B905? Is there anywhere where we use zip without strict requirement?

.pre-commit-config.yaml
```diff
@@ -1,6 +1,6 @@
 repos:
-  - repo: https://github.com/psf/black
-    rev: 24.3.0
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.15.15
     hooks:
-    - id: black
-      language_version: python3
+    - id: ruff
```
milosobral, line 5: You could have the hook directly run fix if possible (I do that on other projects but maybe we don't want it here?)

```
- id: ruff-check
  args: [ --fix ]
```

torch_brain/data/arraydict.py
```diff
@@ -110,8 +109,8 @@ def select_by_mask(self, mask: np.ndarray):
             >>> import numpy as np
 
             >>> units = ArrayDict(
-            ...     unit_id=np.array(["unit01", "unit02"]),
-            ...     brain_region=np.array(["M1", "M1"]),
+            ...     unit_id=["unit01", "unit02"],
+            ...     brain_region=["M1", "M1"],
```
milosobral, line 113: Are these changes related to the typing stuff? (I don't mind packaging them in here if they aren't just curious)

milosobral (commented):
Minor comments otherwise really like moving to ruff

torch_brain/data/arraydict.py
AlexandreAndr: These changes are related to this change https://github.com/neuro-galaxy/temporaldata/pull/123.
Now for quick examples, we should use directly `list` or `ArrayLike` which makes the example more compact

.pre-commit-config.yaml
AlexandreAndr: I am good about adding it! good catch

pyproject.toml
AlexandreAndr: Now we don't use stict with zip, if we do not ignore we have to start using it. I think we could do it I have no preference

milosobral: I think the logic of forcing strict is pretty sound, the default behavior without it is kind of a nightmare

vinamarora8: +1 on enabling B905. Not critical at this moment, and could (or even should) be a separate PR. If so, just open an issue so we remember

vinamarora8 (approved):
LGTM! Assuming most of these are tool created edits; and relying on our test harness :)

pyproject.toml
vinamarora8: Opened an issue: https://github.com/neuro-galaxy/torch_brain/issues/268

PR #263, opened 2026-06-02
merge: Adding a proposal for the glossary

### LINKED ISSUE #264: Add a Glossary to define common language for package

milosobral:
Closes issue #264 

Please feel free to disagree with any definition! Some of them I've had to make an opionated choice so it's expected that you might not agree exactly with the wording or the definition altogether


GLOSSARY.md
```diff
@@ -0,0 +1,143 @@
+# torch_brain Context
+
+Core framework for neural population decoding and processing from electrophysiological recordings.
```
AlexandreAndr: Ideally we would also allow ophys

GLOSSARY.md
```diff
@@ -0,0 +1,143 @@
...
+An event-driven data container for non-uniformly sampled or discrete sequential observations. Defined explicitly by a 1-D array of ascending timestamps paired with arbitrary attributes that must rigidly share the same first dimension as the event axis.
+*Avoid*: Point process, event tracker, sparse tensor, spike list
+
+### Signals & Hardware
+
+**Unit**:
+A source of discrete neural events (spikes). This includes both isolated single neurons and unsorted multi-unit activity (threshold crossings).
+*Avoid*: Neuron, cell
+
+**Channel**:
```
AlexandreAndr, line 55: maybe we can add rois or something standard from ophys

AlexandreAndr: Do we use channel for lfp? units? I have not work much with them

GLOSSARY.md
```diff
@@ -0,0 +1,143 @@
...
+A single 1D time-series stream of data within an array (e.g., a row in an EEG matrix). A channel has a specific modality/type but does not have to be limited to brain data (e.g., stimulus triggers, eye tracking).
+*Avoid*: Stream, track, trace
+
+**Device**:
+The physical hardware used to record the data (e.g., Neuropixels probe, Utah array, EEG cap).
+*Avoid*: Probe, hardware
+
+**Modality**:
```
AlexandreAndr, line 59: Do we want to provide as complete as possible list even if not exhaustive?

GLOSSARY.md
```diff
@@ -0,0 +1,143 @@
...
+The semantic meaning and structural specification of a given Channel, Unit, or output. It dictates the data type (e.g., continuous, discrete), dimensions, and appropriate loss function (e.g., `cursor_velocity_2d`, `lfp`, `eeg`, `spikes`).
+*Avoid*: Feature type, signal type
+
+### Time Segmentation & Manipulation
+
+**Trial**:
+An experimentally defined, behaviorally relevant segment of time (e.g., a stimulus presentation or a reaching movement).
+*Avoid*: Epoch, sweep
+
+**Window**:
```
AlexandreAndr, line 69: here we could avoid slice, interval things that are within our naming

GLOSSARY.md
```diff
@@ -0,0 +1,143 @@
...
+
+**Slicing**:
+The extraction of a specific `[start, end)` time range from a :class:`Data` object, which can be done either deterministically to isolate fixed trial blocks or stochastically to generate random segments for data training diversity.
+*Avoid*: Cropping, chopping, segmenting, patches
+
+**Binning**:
+The process of aggregating sparse discrete events within defined time windows into contiguous, fixed-width intervals to produce a dense, uniformly sampled regular time series array.
+*Avoid*: Downsampling, digitizing, quantizing
+
+### Model Architecture
```
AlexandreAndr: What about embedding?

GLOSSARY.md
milosobral: Not sure about that tbh

GLOSSARY.md
milosobral: I'm not sure about that. The idea here is to define concepts without going into the details and specific  standards of these concepts I think.

GLOSSARY.md
milosobral: I think Interval and Slice are important because they are "real" words where people might expect other meaning. Even for the new concepts that are very specific to this concept, I think it's still valuable to define them concisely here to have a clear definition of their scope (e.g. a RegularTimeSeries should always be for continuous data)

GLOSSARY.md
milosobral: Good point I should add that!

GLOSSARY.md
milosobral: Oh completely misunderstood your message sorry!

GLOSSARY.md
AlexandreAndr: Fair

GLOSSARY.md
AlexandreAndr: sorry my message was not super clear 😅

GLOSSARY.md
vinamarora8: *Any* neural signal. And not just specific to decoding. I would say:

End-to-end ML framework for creating training pipelines and models with neural data.

GLOSSARY.md
AlexandreAndr: I think we can update this later

AlexandreAndr (dismissed):
LGTM but we can wait for @vinamarora8 confirmation too

vinamarora8 (commented):
One of my main concerns here is duplication of definitions. Many of the objects here like RegularTimeseries, Data, etc., have their own docstrings. I think I would prefer to look up the definitions in docs (which are searchable); and I would also expect that the docstrings are always the most *correct* ground truth. 

I understand when it defines "soft" terms, like "Modality", that do not have a definition written in code. But I also feel I did not fully understand the vision of GLOSSARY.md, and would really appreciate some elaboration from your perspective @milosobral!

I know we discussed it in our last meeting; but I have somewhat forgotten about it's complete purpose

milosobral:
@vinamarora8 The main objective of the Glossary is to rigorously define all the terms and concepts that exist in the package. Its goal is not to document any piece of code but to make it clear what the important concpts are and how we decided to refer to them (note that I think that by definition it has to be opinionated). If we have a a complete glossary that we all agree on, things like choosing the names of variables, explaining concepts in the documentation and discussing some issues will become easier because we are all basically guaranteed to be talking about the same things. 

Some examples of terms that could be redundant/unclear and I believe deserve to be defined:
- `recording` vs `session` vs `run`
- `window` vs `time-slice` vs `chunk` vs `patch` 
- `split type` vs `split mode` vs `split strategy`
- I believe the glossary could even be used to define things like the use of `val` vs `valid` as short for validation. Most of the code used to use `valid` but for some reason the last two brainsets (OpenNeuro and Neuroprobe) use `val` now which is not a great user experience. 

I see the GLOSSARY as being a single source of truth for CONCEPTS that we can rely on to decide what to name things and to easily reason about how different pieces of the code interact with each other. 

I do see your point about it basically duplicating some definitions, but I don't see this as an issue. The Glossary definitions should be short and should only define the concepts and have nothing to do with the implementation. If we are changing those definitions often enough that keeping them up to date with the code (or the docstrings of the code) becomes an issue, then I think we've gone wrong somewhere. In general, I expect that most definitions would very rarely change. The glossary would simply expand as we add more features that require more concepts to be defined. 

When it comes to where those definitions should leave, I do think that having a separate file is a good idea as it gives a single point that is easy to reference. I've added it in mardkdown so far for simplicity but it could be anywhere

GLOSSARY.md
```diff
@@ -0,0 +1,156 @@
...
+The total temporal boundaries over which a RegularTimeSeries, IrregularTimeSeries, or Session is defined and holds valid data. Represented as an Interval, it may consist of a single contiguous span or a union of disjoint time ranges.
+*Avoid*: Time span, recording duration, timeline
+
+**Interval**:
+A set of one or more time ranges, each defined by a start time and an end time. May carry additional per-range attributes (e.g. trial metadata). Used to define Domains, represent collections of Trials, and specify Sampling Intervals.
+*Avoid*: Epoch, segment, time range
+
+**Data**:
+A top-level container object that aggregates diverse neural, behavioral, and experimental attributes under a single unified Domain. May nest other Data objects or specialized series containers (RegularTimeSeries, IrregularTimeSeries, ArrayDict), serving as the canonical programmatic representation of a complete Session or a Window sliced from one.
+*Avoid*: Dictionary, record, payload, state map
```
vinamarora8:
```suggestion
**Data**:
A container object that aggregates diverse neural, behavioral, and experimental attributes under a single unified Domain. May nest other Data objects or specialized series containers (RegularTimeSeries, IrregularTimeSeries, Interval, ArrayDict), serving as the canonical programmatic representation of a complete Session or a Window sliced from one.
*Avoid*: Dictionary, record, payload, state map
```

- Can't say top-level since Data objects can be nested (in which case the child Data object is not a top-level container).
- Add `Interval`

vinamarora8: Minor: Also a better ordering would be to put this point either before Interval or after IrregularTimeSeries

GLOSSARY.md
```diff
@@ -0,0 +1,156 @@
...
+
+**ArrayDict**:
+A dictionary-like container of multi-dimensional arrays or tensors that strictly share a uniform first dimension. Permits individual attributes to carry distinct inner dimensions and data types while ensuring rigorous row-wise index synchronization. Used within Data to store per-Unit or per-Channel attributes.
+*Avoid*: Dataframe, table, matrix, structural dictionary
```
vinamarora8, line 39:
```suggestion
**ArrayDict**:
A dictionary-like container of multi-dimensional arrays or tensors that strictly share a uniform first dimension. Permits individual attributes to carry distinct inner dimensions and data types while ensuring rigorous row-wise index synchronization. Typically used within Data to store per-Unit or per-Channel attributes.
*Avoid*: Dataframe, table, matrix, structural dictionary
```

GLOSSARY.md
```diff
@@ -0,0 +1,156 @@
...
+
+**IrregularTimeSeries**:
+An event-driven data container for non-uniformly sampled or discrete sequential observations. Defined explicitly by a 1-D array of ascending timestamps paired with arbitrary attributes sharing the same first dimension. Each instance has its own Domain. Used within Data to represent spike events from Units or irregularly sampled behavioral signals.
+*Avoid*: Point process, event tracker, sparse tensor, spike list
+
+### Signals & Hardware
+
+**Unit**:
+A source of discrete neural events (spikes). This includes both isolated single neurons and unsorted multi-unit activity (threshold crossings). A Session's spike events are stored as an IrregularTimeSeries with a per-spike reference to the originating Unit.
+*Avoid*: Neuron, cell
```
vinamarora8, line 53:
```suggestion
**Unit**:
A putative neuron isolated from a multi-neuron electrophysiological recording, typically using a spike-sorting algorithm. It is a source of discrete neural evens (spikes). A Session's spike events are stored as an IrregularTimeSeries with a per-spike reference to the originating Unit. Metadata about a Session's units is stored as an ArrayDict within the Session's Data object.
*Avoid*: Neuron, cell
```

Making it more precise. Also I don't think we should call a "multi-unit" units as Unit (that would be a misnomer)

GLOSSARY.md
```diff
@@ -0,0 +1,156 @@
...
+
+**Channel**:
+A single 1D time-series stream of data within an array (e.g., a row in an EEG matrix). Each Channel has a specific Modality but does not have to be limited to brain data (e.g., stimulus triggers, eye tracking). Channel data is typically stored as a RegularTimeSeries within a Data object.
+*Avoid*: Stream, track, trace
```
vinamarora8:
```suggestion
A source of 1D time-series data stream within an array (e.g., a row in an EEG matrix). Each Channel has a specific Modality but does not have to be limited to brain data (e.g., stimulus triggers, eye tracking). Channel data is typically stored as a RegularTimeSeries within a Data object.
*Avoid*: Stream, track, trace
```

GLOSSARY.md
```diff
@@ -0,0 +1,156 @@
...
+
+**Device**:
+The physical hardware used to record the data (e.g., Neuropixels probe, Utah array, EEG cap). A Device hosts one or more Units and/or Channels.
+*Avoid*: Probe, hardware
```
vinamarora8: This limits "Device" to be just about the neural recording. Something needs to be added regarding behavioral recording devices (e.g. disk, wheel, camera) eventually.

GLOSSARY.md
```diff
@@ -0,0 +1,156 @@
...
+**Binning**:
+The process of aggregating sparse discrete events within defined time windows into contiguous, fixed-width intervals to produce a dense, uniformly sampled array. Converts an IrregularTimeSeries (e.g., spikes) into a RegularTimeSeries.
+*Avoid*: Downsampling, digitizing, quantizing, pooling
+
+
+### Model Architecture
+
+**Task**:
+The high-level behavioral paradigm or objective the subject is performing during data collection (e.g., "Center-Out Reach", "Sleep").
+*Avoid*: Paradigm, experiment
```
vinamarora8: Hmm this is a bit controversial. I also call "regression-vs-classification-vs-SSL" as defining the tasks. 

Do we need to define it at all?

GLOSSARY.md
```diff
@@ -0,0 +1,156 @@
...
+The process of regularly subdividing a Window along the temporal axis, spatial axis, or both into smaller, uniform sub-windows. Patching is a strategy a Tokenizer may use to decompose signals into Tokens.
+*Avoid*: Cropping, slicing, chopping, segmenting
+
+**Binning**:
+The process of aggregating sparse discrete events within defined time windows into contiguous, fixed-width intervals to produce a dense, uniformly sampled array. Converts an IrregularTimeSeries (e.g., spikes) into a RegularTimeSeries.
+*Avoid*: Downsampling, digitizing, quantizing, pooling
+
+
+### Model Architecture
+
```
vinamarora8: Honestly, I don't know if this is the right time to define these. Many of them are controversial. E.g. what is a task, readout, token, etc. I don't think any sufficiently large ML group would agree to a single definition of either of these at this point in time. 

More importantly, do we even _need_ to define these? Our data pipeline is becoming mature and I see that definitions related to data will stick around. Our model-writing framework, however, is still under heavy development IMO (we have a single model in the library right now). So these seems too early to formalize, and don't have enough weight backing the definitions.

GLOSSARY.md
```diff
@@ -0,0 +1,156 @@
...
+The process of extracting a specific `[start, end)` time range from a Data object, producing a new Data object for that Window. Can be done deterministically (to isolate Trials) or stochastically (to generate random Windows for training diversity).
+*Avoid*: Cropping, chopping, segmenting, patches
+
+**Patching**:
+The process of regularly subdividing a Window along the temporal axis, spatial axis, or both into smaller, uniform sub-windows. Patching is a strategy a Tokenizer may use to decompose signals into Tokens.
+*Avoid*: Cropping, slicing, chopping, segmenting
+
+**Binning**:
+The process of aggregating sparse discrete events within defined time windows into contiguous, fixed-width intervals to produce a dense, uniformly sampled array. Converts an IrregularTimeSeries (e.g., spikes) into a RegularTimeSeries.
+*Avoid*: Downsampling, digitizing, quantizing, pooling
```
vinamarora8: This would be defining processing techniques that exist _independent_ of our library, and not concepts that _we_ introduced. So defining them seems a bit strange to me and doesn't seem like it is our place to do so.

GLOSSARY.md
```diff
@@ -0,0 +1,156 @@
...
+
+### Operations & Evaluation Lifecycle
+
+**Transform**:
+A callable operation that takes a Data object, modifies it (either deterministically or stochastically), and returns it. Transforms are applied after Slicing and before tokenization.
+*Avoid*: Augmentation, preprocessor
+
+**Stitching**:
+The process of aggregating and reconciling model predictions across overlapping Windows to reconstruct a single continuous prediction stream. Required when a Sampler produces Windows with temporal overlap (step < window length).
+*Avoid*: Merging, pooling, aggregating
```
vinamarora8: We need to think about renaming this or something. Another popular use of the word "Stitching" in literature is to use a linear layer to project multiple neural populations into the same latent space (CC @AlexandreAndr). Also, we don't really provide a behavior stitcher at this moment (previous one was removed), so I don't see a need to define.

GLOSSARY.md
```diff
@@ -0,0 +1,156 @@
...
+A batching strategy where variable-length Token sequences are concatenated along the first dimension into a single flat sequence `(Total_T, ...)`, avoiding Padding overhead.
+*Avoid*: Concatenating (as a formal strategy name), flattening
+
+**Split**:
+A specific partition of data designated for a distinct phase of the machine learning lifecycle (strictly `"train"`, `"val"`, or `"test"`).
+*Avoid*: Subset, group, bucket, tier
+
+**Split Strategy**:
+The underlying systemic logic or axis used to segregate time blocks or identifiers into Splits, determined solely by structural boundaries (strictly "intrasession" for slicing a single continuous span, "intersession" for dividing Sessions, or "intersubject" for isolating different biological subjects). A Split Strategy produces one or more Folds.
+*Avoid*: Split type, validation regime, fold method, partition shape
```
vinamarora8, line 101: Isn't `split_type` a term that's already used in our code? Seems more natural to use; but obviously debatable

vinamarora8 (commented):
Milo, thanks for explaining your thought process behind this effort! I do like the direction, but I also want to us be careful and not over-reach. IMO we should limit ourselves to:

- Concepts that _we have introduced_. Concepts that are regularly used in literature, like binning and patching, do not seem to fit this. In these cases, our library is simply implementing ideas defined elsewhere, and the docstring of the implementer is a much better place to describe the details.

- Additionally, we should also limit this effort to concepts that have stood the test of time. Concepts with 1 or 2 instantiations in the library should not be defined until they prove to be stable. I think "tokenizer" would be an example of a concept I think has not proven to be stable yet. 

I would suggest a strategy of starting with a minimal list of "must-define" concepts (like Session) and grow it as we find new concepts introduced by us and that have stuck around for a while.

GLOSSARY.md
AlexandreAndr: Yes I agree on the ambiguity of Stitching at least in the"spike" modality literature

GLOSSARY.md
milosobral: I think we could update the definition here when necessary to reflect a new feature.

GLOSSARY.md
milosobral: Makes sense

GLOSSARY.md
milosobral: Yea I agree, no need to clearly define this at this point. I'll still nuance that I believe the glossary could be useful in fueling discussions about these concepts and how we implement them as we would have to come up with clear definitions. I believe that waiting for the definitions for these concepts to become clear with time is a bit utopic and it will require an effort in terms of design and making some opinionated decisions to end up with a clear definition for some concepts. Long story short, I don't think that everybody will organically land on a general clear definition of a Task for example, and I still believe it would be useful at some point to answer what a task refers to in our codebase, even if some people might disagree.

GLOSSARY.md
milosobral: Same as above, I'll remove those for now

GLOSSARY.md
milosobral: I think both are used in different brainsets. Once we agree on soe of these terms, we will need to do some refactoring to make sure we are coherent everywhere. This is also true for the use of valid vs val for example

vinamarora8: > I think both are used in different brainsets.

Oh noo. Can I convince you to use `split_type` instead of `split_strategy`? Or do you really prefer `split_strategy`?

> This is also true for the use of valid vs val for example.

Yea `val` vs `valid` is so annoying. `val` is by far seems to be the standard term people use when I talk to them so I would be down to switch it all to `val`. (this is just my experience tho)

GLOSSARY.md
vinamarora8: Are we keeping these still? These are standard ML terms so I don't see why we should/need-to define (and I know some people who would really disagree that definition of Embedding).

Maybe we should discuss this over a call

GLOSSARY.md
vinamarora8: > I don't think that everybody will organically land on a general clear definition of a Task for example

Here's how I am viewing it: If things implemented by our library care about there being a clear definition of a Task, then we'll _have_ to define it some day. For example, we all realized that we really need to define what a Session/recording is (and thank you for actually going ahead and writing it!). I imagine the same sort of organic development will work for future concepts as well (?). 

Also, not _everybody_ has to land on a clear general definition. Just us, the people maintaining torch_brain, have to come to a consensus and ensure that different parts of the library align with these definitions.

GLOSSARY.md
vinamarora8: If we know that a current feature implementation has holes, maybe we should defer defining it until we're happy with it? I think defining it would give a signal that we want "Device" to mean this specific thing and adjust future APIs to fit this, rather than the other way around where we come back and change the definition to align with the features.

GLOSSARY.md
milosobral: I'm completely fine with switching everything to `split_type` and `val`

GLOSSARY.md
milosobral: You're right, forgot to remove them sorry

milosobral:
Now in #296

Issue #274, opened 2026-06-04
Fix Pipeline for flint_slutzky_accurate_2012


AlexandreAndr:
In this job https://github.com/neuro-galaxy/torch_brain/actions/runs/26934531852/job/79461070400 This error is triggered:

```
Traceback (most recent call last):
  File "/home/runner/work/_temp/setup-uv-cache/archive-v0/LdSalGzskl8I754H/lib/python3.11/site-packages/urllib3/connection.py", line 204, in _new_conn
    sock = connection.create_connection(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/_temp/setup-uv-cache/archive-v0/LdSalGzskl8I754H/lib/python3.11/site-packages/urllib3/util/connection.py", line 85, in create_connection
    raise err
  File "/home/runner/work/_temp/setup-uv-cache/archive-v0/LdSalGzskl8I754H/lib/python3.11/site-packages/urllib3/util/connection.py", line 73, in create_connection
    sock.connect(sa)
TimeoutError: [Errno 110] Connection timed out

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/runner/work/_temp/setup-uv-cache/archive-v0/LdSalGzskl8I754H/lib/python3.11/site-packages/urllib3/connectionpool.py", line 788, in urlopen
    response = self._make_request(
               ^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/_temp/setup-uv-cache/archive-v0/LdSalGzskl8I754H/lib/python3.11/site-packages/urllib3/connectionpool.py", line 488, in _make_request
    raise new_e
  File "/home/runner/work/_temp/setup-uv-cache/archive-v0/LdSalGzskl8I754H/lib/python3.11/site-packages/urllib3/connectionpool.py", line 464, in _make_request
    self._validate_conn(conn)
  File "/home/runner/work/_temp/setup-uv-cache/archive-v0/LdSalGzskl8I754H/lib/python3.11/site-packages/urllib3/connectionpool.py", line 1106, in _validate_conn
    conn.connect()
  File "/home/runner/work/_temp/setup-uv-cache/archive-v0/LdSalGzskl8I754H/lib/python3.11/site-packages/urllib3/connection.py", line 759, in connect
    self.sock = sock = self._new_conn()
                       ^^^^^^^^^^^^^^^^
[...]

PR #279, opened 2026-06-05
ci: Allow Pipeline Steps to Continue on Failure

AlexandreAndr:
- Adds continue-on-error: true to each brainsets prepare step so a single dataset download failure no longer blocks the rest of the pipeline
- Moves the brainsets config set call to a single shared step, removing the duplicated invocation that appeared in every download step
- Re-enables vollan_moser_alternating_2025 (navigation and sleep sessions) which were previously commented out due to an SSL cert issue with data-proxy.ebrains.eu

### DIFF
--- a/.github/workflows/pipeline_testing.yml
+++ b/.github/workflows/pipeline_testing.yml
@@ -29,60 +29,50 @@ jobs:
     - name: Install the project
       run: uv sync --all-extras
 
+    - name: Configure brainsets
+      run: uv run brainsets config set --raw-dir data/raw --processed-dir data/processed
+
     - name: Download pei_pandarinath_nlb_2021
-      run: |
-        uv run brainsets config set --raw-dir data/raw --processed-dir data/processed
-        uv run brainsets prepare pei_pandarinath_nlb_2021
+      continue-on-error: true
+      run: uv run brainsets prepare pei_pandarinath_nlb_2021
 
     - name: Download perich_miller_population_2018
-      run: |
-        uv run brainsets config set --raw-dir data/raw --processed-dir data/processed
-        uv run brainsets prepare perich_miller_population_2018 -s c_20131219_center_out_reaching
+      continue-on-error: true
+      run: uv run brainsets prepare perich_miller_population_2018 -s c_20131219_center_out_reaching
 
     - name: Download allen_visual_coding_ophys_2016
-      run: |
-        uv run brainsets config set --raw-dir data/raw --processed-dir data/processed
-        uv run brainsets prepare allen_visual_coding_ophys_2016 -s 717913184
+      continue-on-error: true
+      run: uv run brainsets prepare allen_visual_coding_ophys_2016 -s 717913184
 
     - name: Download churchland_shenoy_neural_2012
-      run: |
-        uv run brainsets config set --raw-dir data/raw --processed-dir data/processed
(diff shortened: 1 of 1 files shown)


AlexandreAndr:
Related to issue #274

.github/workflows/pipeline_testing.yml
```diff
@@ -29,60 +29,50 @@ jobs:
...
-        uv run brainsets config set --raw-dir data/raw --processed-dir data/processed
-        uv run brainsets prepare kemp_sleep_edf_2013 -s sleep_cassette_SC4002E0-PSG
+      run: uv run brainsets prepare kemp_sleep_edf_2013 -s sleep_cassette_SC4002E0-PSG
 
     # TODO: SSL cert issue with data-proxy.ebrains.eu, re-enable when fixed
-    # - name: Download vollan_moser_alternating_2025 (navigation)
-    #   run: |
-    #     uv run brainsets config set --raw-dir data/raw --processed-dir data/processed
-    #     uv run brainsets prepare vollan_moser_alternating_2025 -s of_24365_2
+    - name: Download vollan_moser_alternating_2025 (navigation)
```
milosobral, line 68: You didn't add continue on error for the two vollan_moser steps, is that on purpose?

milosobral (commented):
Minor coment but otherwise looks good!

Too bad it worked that time so we can't really see what it woul look like when one of the steps fail 😁. Any idea if we could force one pipeline to fail to get an idea of that?

vinamarora8:
> Any idea if we could force one pipeline to fail to get an idea of that?

We can keep retriggering until it fails :3

AlexandreAndr:
Yes I just relaunched the job and let's see if it fails!

.github/workflows/pipeline_testing.yml
AlexandreAndr: Good catch I'll add it!

.github/workflows/pipeline_testing.yml
```diff
@@ -29,60 +29,87 @@ jobs:
...
+          "${{ steps.kochi_visualnaming_ds006914.outcome }}"
+          "${{ steps.shirazi_hbnr1_ds005505.outcome }}"
+        )
+        failed=0
+        for outcome in "${outcomes[@]}"; do
+          if [ "$outcome" == "failure" ]; then
+            failed=1
+          fi
+        done
+        exit $failed
```
vinamarora8: Something like this might be better. This should automatically check all steps (don't have to list manually), and print the names of the steps that did fail.

(This is from an AI chatbot, so might have issues, but I hope you get the idea I am trying to convey)

```suggestion
  - name: Check all pipelines passed
    if: always()
    env:
      STEPS_JSON: ${{ toJSON(steps) }}
    run: |
      failed=$(echo "$STEPS_JSON" | jq -r 'to_entries[] | select(.value.outcome == "failure") | .key')
      if [ -n "$failed" ]; then
        echo "The following pipelines failed:"
        echo "$failed" | sed 's/^/  - /'
        exit 1
      fi
      echo "All pipelines passed."
```

AlexandreAndr: totally agree I'll test your update

vinamarora8 (commented):
LGTM!

PR #281, opened 2026-06-06
docs: improve torch_brain.data guide

### LINKED ISSUE #253: Docs: New flow for `torch_brain.data` user guide

vinamarora8:
Addressing the remaining concerns in #253.

Reword the `torch_brain.data` user guide from having a conversational tone to an assertive tone. This is following [milo's suggestion](https://github.com/neuro-galaxy/torch_brain/pull/254#pullrequestreview-4428485059), which stems from how PyTorch writes its guides.

See the docs [here](https://torch-brain--281.org.readthedocs.build/en/281/)

### DIFF
--- a/docs/source/guides/data/domain.rst
+++ b/docs/source/guides/data/domain.rst
@@ -132,7 +132,7 @@ while ``behavior`` spans only :math:`[1s, 9s)`, so the union is :math:`[0s, 10s)
 Why domains matter
 ------------------
 
-When you train a model in **torch_brian**, :ref:`Samplers <samplers_ref>` use
+When you train a model in **torch_brain**, :ref:`Samplers <samplers_ref>` use
 each recording's domain to decide *where* they can draw time-windows from.
 Sampling from time periods where the data is absent or corrupted should be
 avoided, so removing those periods from the domain is good practice. In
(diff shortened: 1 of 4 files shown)


docs/source/guides/data/domain.rst
```diff
@@ -135,8 +132,9 @@ spans only :math:`[1, 9)`, so the union is :math:`[0, 10)`:
 Why domains matter
 ------------------
 
-When you train a model in TorchBrain, :ref:`Samplers <samplers_ref>` look at
```
milosobral, line 138: Here, you could maybe link to the sampler guide instead of the API reference (or both)

milosobral (approved):
Great changes! I think these make it easier to understand the concepts efficiently and to get the important information accross. 

Small detail: I feel like separating the GappyRegularTimeSeries page miught make it seem like it conceptually is on a different level to the other objects when it isn't. What do you think about including it in the dat_objects_intro instead of it being its own page?

vinamarora8:
> Great changes!

Thanks!

> What do you think about including it in the dat_objects_intro instead of it being its own page?

I would like it to be its own page for the moment. It talks about details (like how domain edge cases are handled) that don't seem easy enough to fit in the first user guide of the package IMO. (Note: Domain itself is introduced in the third guide)

PR #288, opened 2026-06-08
Limiting Benchmark Runs

### LINKED ISSUE #285: Benchmark workflow should not trigger on every PR
It is a little annoying when the benchmark scores shows up on completely unrelated PRs. Perhaps it should trigger only if changes are made to the following?
- Anything in `torch_brain/data/*`
- `pyproject.toml` (since dependency changes can change speed of certain operations)

milosobral:
Added paths to the benchmarks workflow to ensure it triggers on changes to relevant files in the data and scripts directories, as well as the workflow file itself. Closes #285 

I've created a non-benchmarkable PR #289 and a benchmarkable PR #290 to make sure it behaves properly and it all looks good.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -3,6 +3,11 @@ name: Benchmarks
 on:
   pull_request:
     types: [opened, reopened, synchronize]
+    paths:
+      - torch_brain/data/**
+      - pyproject.toml
+      - scripts/data_benchmarks/**
+      - .github/workflows/benchmarks.yml
 
 permissions:
   contents: read


.github/workflows/benchmarks.yml
```diff
@@ -3,6 +3,11 @@ name: Benchmarks
 on:
   pull_request:
     types: [opened, reopened, synchronize]
+    paths:
+      - torch_brain/data/**
+      - pyproject.toml
```
AlexandreAndr, line 8: Why including pyproject.toml?

AlexandreAndr (approved):
Just one curious out of curiosity, otherwise LGTM

.github/workflows/benchmarks.yml
milosobral: I think if the dependencies change, it could technically impact the performance (Numpy or h5py version for example)

PR #299, opened 2026-06-11
feat: Add `delete_nested_attribute` in torch_brain.data`

### LINKED ISSUE #295: Add a `delete_nested_attribute` in `torch_brain.data` for `Data` object
For Data we altready provide these 3 functions:
- `get_nested_attribute()`
- `set_nested_attribute()`
- `has_nested_attribute()`
    
In the same spirit, it would be nice to have a function to delete a nested attribute

AlexandreAndr:
## Summary

- Adds `Data.delete_nested_attribute(path)` symmetric with the existing `get`/`set`/`has` methods
- Deletion guards live on `__delattr__` of each class (`Data`, `ArrayDict`, `IrregularTimeSeries`, `Interval`), not on the API method, so protection applies regardless of how `delattr` is called:
  - `IrregularTimeSeries.timestamps` and `Interval.start`/`end` always blocked
  - Other registered timekeys can be deleted and are auto-deregistered from `_timekeys`
- `_resolve_to_parent` helper refactors shared path-walking logic from `get`/`set`/`delete`
- Fixes latent bug: `LazyIrregularTimeSeries` was not explicitly imported in `data.py`

See docs [here](https://torch-brain--299.org.readthedocs.build/en/299/generated/api/autosummary/torch_brain.data.Data.html#torch_brain.data.Data)

### DIFF
--- a/tests/data/test_data.py
+++ b/tests/data/test_data.py
@@ -36,8 +36,8 @@ def test_data():
         session_id="session_0",
         domain=Interval.from_list([(0, 3)]),
         spikes=IrregularTimeSeries(
-            timestamps=np.array([0.1, 0.2, 0.3, 2.1, 2.2, 2.3]),
-            unit_index=np.array([0, 0, 1, 0, 1, 2]),
+            timestamps=[0.1, 0.2, 0.3, 2.1, 2.2, 2.3],
+            unit_index=[0, 0, 1, 0, 1, 2],
             waveforms=np.zeros((6, 48)),
             domain="auto",
         ),
@@ -46,14 +46,14 @@ def test_data():
             sampling_rate=250.0,
         ),
         units=ArrayDict(
-            id=np.array(["unit_0", "unit_1", "unit_2"]),
-            brain_region=np.array(["M1", "M1", "PMd"]),
+            id=["unit_0", "unit_1", "unit_2"],
+            brain_region=["M1", "M1", "PMd"],
         ),
         trials=Interval(
-            start=np.array([0, 1, 2]),
-            end=np.array([1, 2, 3]),
-            go_cue_time=np.array([0.5, 1.5, 2.5]),
-            drifting_gratings_dir=np.array([0, 45, 90]),
+            start=[0, 1, 2],
+            end=[1, 2, 3],
+            go_cue_time=[0.5, 1.5, 2.5],
+            drifting_gratings_dir=[0, 45, 90],
         ),
         drifting_gratings_imgs=np.zeros((8, 3, 32, 32)),
     )
@@ -89,10 +89,10 @@ def test_data_copy():
     data = Data(
         session_id="session_0",
         domain=Interval.from_list([(0, 3)]),
-        some_numpy_array=np.array([1, 2, 3]),
+        some_numpy_array=[1, 2, 3],
         spikes=IrregularTimeSeries(
-            timestamps=np.array([0.1, 0.2, 0.3, 2.1, 2.2, 2.3]),
-            unit_index=np.array([0, 0, 1, 0, 1, 2]),
+            timestamps=[0.1, 0.2, 0.3, 2.1, 2.2, 2.3],
+            unit_index=[0, 0, 1, 0, 1, 2],
             waveforms=np.zeros((6, 48)),
             domain="auto",
         ),
@@ -101,14 +101,14 @@ def test_data_copy():
             sampling_rate=250.0,
         ),
         units=ArrayDict(
-            id=np.array(["unit_0", "unit_1", "unit_2"]),
-            brain_region=np.array(["M1", "M1", "PMd"]),
+            id=["unit_0", "unit_1", "unit_2"],
+            brain_region=["M1", "M1", "PMd"],
         ),
         trials=Interval(
-            start=np.array([0, 1, 2]),
-            end=np.array([1, 2, 3]),
-            go_cue_time=np.array([0.5, 1.5, 2.5]),
-            drifting_gratings_dir=np.array([0, 45, 90]),
+            start=[0, 1, 2],
+            end=[1, 2, 3],
+            go_cue_time=[0.5, 1.5, 2.5],
+            drifting_gratings_dir=[0, 45, 90],
         ),
         drifting_gratings_imgs=np.zeros((8, 3, 32, 32)),
     )
@@ -142,10 +142,10 @@ def test_lazy_data_copy(test_filepath):
     data = Data(
         session_id="session_0",
         domain=Interval.from_list([(0, 3)]),
-        some_numpy_array=np.array([1, 2, 3]),
+        some_numpy_array=[1, 2, 3],
         spikes=IrregularTimeSeries(
-            timestamps=np.array([0.1, 0.2, 0.3, 2.1, 2.2, 2.3]),
-            unit_index=np.array([0, 0, 1, 0, 1, 2]),
+            timestamps=[0.1, 0.2, 0.3, 2.1, 2.2, 2.3],
+            unit_index=[0, 0, 1, 0, 1, 2],
             waveforms=np.zeros((6, 48)),
             domain="auto",
         ),
@@ -154,14 +154,14 @@ def test_lazy_data_copy(test_filepath):
             sampling_rate=250.0,
         ),
         units=ArrayDict(
-            id=np.array(["unit_0", "unit_1", "unit_2"]),
-            brain_region=np.array(["M1", "M1", "PMd"]),
+            id=["unit_0", "unit_1", "unit_2"],
+            brain_region=["M1", "M1", "PMd"],
         ),
         trials=Interval(
-            start=np.array([0, 1, 2]),
-            end=np.array([1, 2, 3]),
-            go_cue_time=np.array([0.5, 1.5, 2.5]),
-            drifting_gratings_dir=np.array([0, 45, 90]),
+            start=[0, 1, 2],
+            end=[1, 2, 3],
+            go_cue_time=[0.5, 1.5, 2.5],
+            drifting_gratings_dir=[0, 45, 90],
         ),
         drifting_gratings_imgs=np.zeros((8, 3, 32, 32)),
     )
@@ -210,10 +210,10 @@ def test_data_absolute_start(test_filepath):
     data = Data(
         session_id="session_0",
         domain=Interval.from_list([(0, 3)]),
-        some_numpy_array=np.array([1, 2, 3]),
+        some_numpy_array=[1, 2, 3],
         spikes=IrregularTimeSeries(
-            timestamps=np.array([0.1, 0.2, 0.3, 2.1, 2.2, 2.3]),
-            unit_index=np.array([0, 0, 1, 0, 1, 2]),
+            timestamps=[0.1, 0.2, 0.3, 2.1, 2.2, 2.3],
+            unit_index=[0, 0, 1, 0, 1, 2],
             waveforms=np.zeros((6, 48)),
             domain="auto",
         ),
@@ -222,14 +222,14 @@ def test_data_absolute_start(test_filepath):
             sampling_rate=250.0,
         ),
         units=ArrayDict(
-            id=np.array(["unit_0", "unit_1", "unit_2"]),
-            brain_region=np.array(["M1", "M1", "PMd"]),
+            id=["unit_0", "unit_1", "unit_2"],
+            brain_region=["M1", "M1", "PMd"],
         ),
         trials=Interval(
-            start=np.array([0, 1, 2]),
-            end=np.array([1, 2, 3]),
-            go_cue_time=np.array([0.5, 1.5, 2.5]),
-            drifting_gratings_dir=np.array([0, 45, 90]),
+            start=[0, 1, 2],
+            end=[1, 2, 3],
+            go_cue_time=[0.5, 1.5, 2.5],
+            drifting_gratings_dir=[0, 45, 90],
         ),
         drifting_gratings_imgs=np.zeros((8, 3, 32, 32)),
     )
@@ -279,8 +279,8 @@ def test_timeless_data(test_filepath):
     data = Data(
         subject=subject,
         spikes=IrregularTimeSeries(
-            timestamps=np.array([0.1, 0.2, 0.3, 2.1, 2.2, 2.3]),
-            unit_index=np.array([0, 0, 1, 0, 1, 2]),
+            timestamps=[0.1, 0.2, 0.3, 2.1, 2.2, 2.3],
+            unit_index=[0, 0, 1, 0, 1, 2],
             waveforms=np.zeros((6, 48)),
             domain="auto",
         ),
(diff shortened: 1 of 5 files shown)


AlexandreAndr:
Close #295

AlexandreAndr:
@milosobral I think you are the one that implemented `has_nested_ attribute `. 
I remember you had a point to use `__dict__`  rather than `getattr`, I tried to mention it in the info block of the docstring, is the explanation of the docstring is accurate

torch_brain/data/arraydict.py
```diff
@@ -83,6 +83,13 @@ def __setattr__(self, name, value):
                 )
         super().__setattr__(name, value)
 
+    def __delattr__(self, name):
+        if name.startswith("_"):
+            raise AttributeError(
+                f"Cannot delete private attribute '{name}' from {self.__class__.__name__}."
+            )
+        super().__delattr__(name)
+
```
vinamarora8: @AlexandreAndr is the motivation to add this just to avoid things like `data.delete_nested_attribute("a.b._private")`? Or you think there are other cases where preventing deletion of private attributes required? Or is it just a good hygiene thing to not let people delete private attribs?

torch_brain/data/data.py
```diff
@@ -616,6 +664,38 @@ def has_nested_attribute(self, path: str) -> bool:
 
         return True
```
vinamarora8, line 618: Would be great if you also align this to look be like the others:


```suggestion
try:
    obj, attr = self._resolve_to_parent(path)
    getattr(obj, attr)
    return True
except AttibuteError:
    return False
```

torch_brain/data/data.py
```diff
@@ -550,34 +557,62 @@ def __contains__(self, key: str) -> bool:
...
+        return obj, components[-1]
+
+    def get_nested_attribute(self, path: str) -> Any:
+        r"""Return the value of a nested attribute specified by its dot-separated path.
+
+        .. code-block:: python
+
+            for attr in ["timestamps", "unit_index", "waveforms"]:
+                print(attr, data.get_nested_attribute(f"spikes.{attr}"))
+
```
vinamarora8: These should be in an `Examples:` section no? Like:
```python
Examples:
    >>> code-code-code
```

torch_brain/data/arraydict.py
AlexandreAndr: Good point, I think this is a hedge case that would impact only 1% of the users. 
Those users probably know what they are doing if they are deleting private attributes so we could still allow it.
My intuition is that in principle no user even expert ones should delete/set/modify a private attribute as it is suposed to be reserved for internal logic of the dataobject and only managed by the class itself.

For example, imo we should not allow to touch manually `_lazy_ops` in `LazyArrayDict` or `_timekeys` in irregular ts

torch_brain/data/data.py
AlexandreAndr: I agree, I think it is a bit complicated because of this 
```
 Uses ``__dict__`` rather than ``getattr`` at each level of the path
            traversal to avoid two side-effects:
```

We do not ask the same question between set/delete/update where we need to touch the object and has attribute where we just want to know its existance

torch_brain/data/data.py
AlexandreAndr: It seems that some methods use `.. code-block:: python` like `save()`
Some others use `Examples:` like `init()`
I don't know if there is an difference such as one is checked in the pipeline (e.g., check the example could really run)? 
I can change it if you think the `Examples:` is a better convention

torch_brain/data/arraydict.py
vinamarora8: IMO making attributes private is already an indication to users to not meddle with them, unless they really know what they're doing. Enforcing it in code like this is not Pythonic in my view.

torch_brain/data/data.py
```diff
@@ -586,22 +621,35 @@ def set_nested_attribute(self, path: str, value: Any) -> Data:
...
-            try:
-                obj = getattr(obj, c)
-            except AttributeError as err:
-                raise AttributeError(
-                    f"Could not resolve {path} in data (specifically, at level {c})"
-                ) from err
-
-        setattr(obj, components[-1], value)
+        obj, attr = self._resolve_to_parent(path)
+        setattr(obj, attr, value)
```
milosobral, line 624: Should this maybe have the same try-catch as get_attr?

torch_brain/data/data.py
```diff
@@ -616,6 +664,38 @@ def has_nested_attribute(self, path: str) -> bool:
...
+
+        Args:
+            path: Nested attribute path (dot-separated).
+
+        Raises:
+            AttributeError: If any component of the path cannot be resolved, or
+                if the target attribute is protected.
+        """
+        obj, attr = self._resolve_to_parent(path)
+        delattr(obj, attr)
```
milosobral: Same here

torch_brain/data/irregular_ts.py
```diff
@@ -136,6 +136,16 @@ def register_timekey(self, timekey: str):
...
 
+    def __delattr__(self, name):
+        if name == "timestamps":
+            raise AttributeError(
+                f"Cannot delete 'timestamps' from {self.__class__.__name__}."
+            )
+        is_timekey = name in self._timekeys
+        super().__delattr__(name)
+        if is_timekey:
+            self._timekeys = [k for k in self._timekeys if k != name]
```
milosobral: I think the logic here is a bit convoluted and could be replaced by: 
`if name in self._timekeys: self._timekeys.remove(name)`

milosobral (commented):
Left a few comments about implementation but overall like the idea! 
On the private attrivutes discussion, I agree with the idea that the underscore should be enough deterent and that there is no need to hard enforce that.

torch_brain/data/data.py
AlexandreAndr: For setting the attribute I think that it depends on the design you want to make.
Let's imagine the last attribute that you want to set does not exist
- If you use  try-catch you won't allow to set this attribute because it is not yet in the data object
- If you keep it like the current PR state you will create and set this new attribute

Would you prefer to restrict set_nested_attribute to overiding attribute?

AlexandreAndr: The current behvior of set_nested_attribute is to allow creating new attributes but we can update it in this PR

torch_brain/data/data.py
AlexandreAndr: I think in this case it make sense because there will be an error if we try to delete to an attribute that does not exist. I update it

AlexandreAndr: great catch!

torch_brain/data/irregular_ts.py
AlexandreAndr: Done

torch_brain/data/arraydict.py
AlexandreAndr: I remove it

torch_brain/data/data.py
milosobral: Actually I think the current behavior makes more sense so that is fine

torch_brain/data/arraydict.py
```diff
@@ -382,7 +382,8 @@ def __getattribute__(self, name):
                 if all_loaded:
                     self.__class__ = ArrayDict
                     # delete special private attributes
-                    del self._lazy_ops, self._unicode_keys
+                    object.__delattr__(self, "_lazy_ops")
+                    object.__delattr__(self, "_unicode_keys")
                 return out
 
         return super().__getattribute__(name)
```
vinamarora8, line 388: This change is not needed, now that `__delattr__` is unchanged right?

torch_brain/data/data.py
vinamarora8: `Examples:` should be the _correct_ sphinx convention to use in places where we want show an example for how to use a particular API. Right? "Example" is in its name.

I would think `.. code-block::` would be for when we want to just write a block of code, that is not an Example. We can change other code-blocks to Examples.

torch_brain/data/data.py
vinamarora8: Ok makes sense! I hadn't read the docstring.

torch_brain/data/interval.py
```diff
@@ -912,7 +921,8 @@ def __getattribute__(self, name):
                 )
                 if all_loaded:
                     self.__class__ = Interval
-                    del self._lazy_ops, self._unicode_keys
+                    object.__delattr__(self, "_lazy_ops")
+                    object.__delattr__(self, "_unicode_keys")
 
                 return out
```
vinamarora8, line 926: Change not needed right? We should rely on current's classes `__delattr__`

torch_brain/data/arraydict.py
AlexandreAndr: Not it is not, I was thinking that it might be a good design for internal deletion of attribute but if you want I can revert back

torch_brain/data/data.py
AlexandreAndr: After digging a bit, I can confirm that using `Examples:` automatically test the code in the example. It can be annoying because sometimes you do just want a quick few line example without import and defining objects. I think this explained why most docstring were using `.. code-block::` to avoid triggering tests.

To fix that we have several options:
- Keep as is but people might be confused of what to use between `Examples:` and `.. code-block::` + we do not respect the sphinx convention
-Convert all of the to `Examples:` with ` >>>` inside but we need to use `# doctest: +SKIP` to avoid trigering the test for them

I have gone with option 2 for the moment but we can iterate on this

AlexandreAndr: [Here](https://torch-brain--299.org.readthedocs.build/en/299/generated/api/autosummary/torch_brain.data.Data.html#torch_brain.data.Data) is the docs with the change. I can also find runnable and better example for some of them if you think it can/should be improved

torch_brain/data/arraydict.py
vinamarora8: I think it is better to rely on `delattr()` (or equivalently `del xyz`) which effectively calls `self.__delattr__()`. If we use `object.__delattr__()`, then we are always bypassing the default inheritance structure and straight-away going to the top-level parent (`object`). This might seem as a non-issue right now, but I don't see a reason to break the default inheritance structure at this moment. It might save us in the future when we do actually want some custom functionality in `self.__delattr__()`? What do you think?

vinamarora8: Also, we are relying on `setattr`, `getattr`, and (sometimes) `delattr` in other places inside `torch_brain.data`, so let's keep with the same pattern? 

At some places we do have `object.__delattr__()`, though I think @milosobral added it for 1) performance reasons (avoids a nested function call), and 2) to explicitly avoid some logic in the object's `delattr`. IIRC, these _should_ all be in our lazy objects only.

It is really not a big deal, so I can also live with `object.__delattr__()` in your situation.

tests/data/test_data.py
```diff
@@ -537,12 +537,130 @@ def test_data_has_nested_attribute_lazy(test_filepath):
...
+                id=["unit_0", "unit_1", "unit_2"],
+                brain_region=["M1", "M1", "PMd"],
+            ),
+        )
+
+    # --- basic deletion ---
+
+    def test_delete_top_level(self, data):
+        data.delete_nested_attribute("session_id")
+        assert "session_id" not in data.keys()
```
vinamarora8: Would be amazing to also have a `pytest.raises(AttributeError, ...)` check when `data.session_id` is accessed. (Since in weird situation it could that `data.keys()` does not return the attribute, but it actually exists in the data, e.g. if we change the implementation of `.keys()` in the future.)

torch_brain/data/arraydict.py
AlexandreAndr: okay let's keep del xyz for now and we can always revisit when needed after!

tests/data/test_data.py
AlexandreAndr: good catch just updated it!

torch_brain/data/data.py
AlexandreAndr: @vinamarora8 What do you think about this?

vinamarora8: Yes, I meant to say use `Examples:` and SKIP doctest if necessary! Sorry, I should have explicitly included all those details!
It would be great to eventually have most examples actually be runnable, though it is a separate effort and can happen after this PR!

torch_brain/data/data.py
```diff
@@ -616,6 +654,41 @@ def has_nested_attribute(self, path: str) -> bool:
...
+        .. warning::
+            Certain attributes are protected and cannot be deleted:
+
+            - ``timestamps`` on :obj:`IrregularTimeSeries` (structurally required).
+            - ``start`` and ``end`` on :obj:`Interval` (structurally required).
+            - Other registered timekeys on :obj:`IrregularTimeSeries` can be deleted;
+              they are automatically deregistered from the timekeys list so that
+              slicing remains correct.
+            - Other registered timekeys on :obj:`Interval` can be deleted and are
+              automatically deregistered.
```
vinamarora8: Nit: Too much detail IMO. If you like these and want to keep them, I would suggest atleast removing the last two points which are just saying "hey things will happen like you already expect" :3 "Warnings" should be used for things that are surprising (again, IMO).

tests/data/test_data.py
```diff
@@ -537,12 +537,131 @@ def test_data_has_nested_attribute_lazy(test_filepath):
...
+    def test_delete_nested_column_arraydict(self, data):
+        data.delete_nested_attribute("units.brain_region")
+        assert "brain_region" not in data.units.keys()
+        assert "id" in data.units.keys()
+
+    def test_delete_non_timekey_interval_column(self, data):
+        data.delete_nested_attribute("trials.condition")
+        assert "condition" not in data.trials.keys()
+        assert "start" in data.trials.keys()
+        assert "end" in data.trials.keys()
```
vinamarora8, line 596: Can u also add AttributeError checks to these three tests? We want to test for _both_: deleted attributes should not be present in keys(), and they also should not exist within the object (i.e. should give AttributeError on access).

vinamarora8 (commented):
Hey Alex, thanks for the previous edits! Sorry I'm being a pain in the ass with these nits, but I am thinking some of these nits are important/easier to handle at this moment than putting off to later.

torch_brain/data/data.py
AlexandreAndr: I think we can remove it, I think you are right all those are not suprising and not informative the normal user should not think about this

tests/data/test_data.py
AlexandreAndr: good points thanks for checking the tests!
Just added them!

torch_brain/data/data.py
```diff
@@ -430,21 +424,13 @@ def load(cls, path: Path | str, lazy: bool = True) -> Data:
...
-            # non-lazy (no close needed)
-            data = Data.load("data.h5", lazy=False)
+        Examples:
+            >>> from torch_brain.data import Data  # doctest: +SKIP
+            >>> with Data.load("data.h5") as data:  # doctest: +SKIP
+            ...     ...
+            >>> data = Data.load("data.h5")  # doctest: +SKIP
+            >>> data.close()  # doctest: +SKIP
+            >>> data = Data.load("data.h5", lazy=False)  # doctest: +SKIP
         """
```
vinamarora8, line 448: comments got removed :( 
(like `# lazy with context manager (recommended)`)
can u put them back pls?

vinamarora8 (commented):
final comment

torch_brain/data/data.py
AlexandreAndr: good catch! just did

vinamarora8 (approved):
LGTM!

Issue #286, opened 2026-06-06
Usage of `serialize_fn_map` can be hidden/simplified


vinamarora8:
1. `torch_brain.data.serialize_fn_map` is actually a constant, so should be using capital letters for its name. Maybe something like `DEFAULT_SERIALIZE_FN_MAP` seems better.
2. The only place we use it is when saving Data, i.e. `Data.save()` or `Data.to_hdf5()`. We could just pass it as a default argument in there, so most end-users never have to interact with it.

PR #300, opened 2026-06-15
docs: module API reference organization

### LINKED ISSUE #236: Module docstrings
Add/fix module-level docstrings after package merge

### LINKED ISSUE #243: Docs: Fix dunder methods not showing up for `torch_brain.data` classes

### LINKED ISSUE #245: Docs: References to `temporaldata` and `brainsets`
Remove references to `temporaldata` and `brainsets` packages in `:obj:` calls.

vinamarora8:
Organize and add descriptions for the module-level docstring pages of 4 modules: `data`, `datasets`, `pipeline`, and `batching`.

Closes #245, #236 and its sub-issues


torch_brain/data/__init__.py
```diff
@@ -1,5 +1,6 @@
-"""
-**User guide.** See :ref:`data_guide` section for further details.
+"""Core data structures for representing neural and behavioral data.
+
+**User guide.** See :ref:`data_guide` guide for further details.
```
milosobral: Nitpick but I think the bold **User Guide** here is a bit confusing as it feels like it should either be a title or a link. I think you can simply remove it and just keep the "See data_guide for further details." with the link.

torch_brain/data/__init__.py
```diff
@@ -39,7 +40,42 @@
...
+                "SubjectDescription",
+            ],
+        },
+        {
+            "title": "Utility Functions",
+            "autosummary": ["concat"],
+        },
+        {
+            "title": "Constants",
+            "autosummary": ["serialize_fn_map"],
```
milosobral, line 78: This one does not properly link to the page for `serialize_fn_map`

torch_brain/datasets/__init__.py
```diff
@@ -84,24 +47,55 @@
...
+  datasets through a unified interface.
+- :ref:`Mixin <datasets_ref-mixins>` classes are provided to add
+  modality-specific functionalities to the Dataset classes.
+
+
+**Dataset:** torch_brain's :class:`Dataset` class (and its sub-classes) allow
+you to sample *time-slices* of your data. This is a major deviation from the
+standard :class:`torch.utils.data.Dataset`, which is indexed by integers. To
+achieve arbitrary time-slice based access, our Dataset class is indexed by
+three things:
```
milosobral: Very nitpick but don't love the word "things" here, what about "parameters" or "arguments"?

torch_brain/pipeline/__init__.py
```diff
@@ -1,3 +1,8 @@
+"""Tools for defining pipelines that create brainsets.
+
+**User guide.** See :ref:`brainsets_guide` guide for more information.
```
milosobral: Same comment here about "User Guide"

milosobral (commented):
Some very minor comments but otherwise looks good

torch_brain/data/__init__.py
vinamarora8: Yeah this is a constant and sphinx doesn't handle constants well (I don't think anyone can, since constants inherently don't have docstrings). 

I've had an idea for a while to hide this constant all together. And so don't wanna spend time fixing this issue in its doc.

vinamarora8: I've mentioned a potential way to hide it in #286

AlexandreAndr:
Sorry unrelated to the PR but I am thinking about the job pipeline-testing that is failing do we want to trigger the job when we touch the files related to pipeline like here with the docs? It is a more general question of now that we have all packages do we want to rethink the test triggers or maybe it is not a priority?

torch_brain/datasets/__init__.py
```diff
@@ -84,24 +47,55 @@
...
+
```
AlexandreAndr, line 66: I am wondering if using a block of code would catch better a user attention?
```python
DatasetIndex(
    recording_id=... , # The recording id from which you want the slice
    start=... , # Start time of the slice
    end=... , # End time of the slice

)
```
what do you think?

AlexandreAndr (commented):
One minor comment otherwise LGTM

torch_brain/datasets/__init__.py
vinamarora8: good idea. done!

vinamarora8:
> Sorry unrelated to the PR but I am thinking about the job pipeline-testing that is failing do we want to trigger the job when we touch the files related to pipeline like here with the docs? It is a more general question of now that we have all packages do we want to rethink the test triggers or maybe it is not a priority?

@AlexandreAndr can u open an issue to discuss this please. We do need to address it somehow.

milosobral (approved):
LGTM

PR #301, opened 2026-06-15
cleanup!: Remove `torch_brain.nn.Embedding`

### LINKED ISSUE #294: Remove `nn.Embedding`
`torch_brain.nn.Embedding` is a minor change over `torch.nn.Embedding`. The extra feature is that it allows for setting weight initialization scale easily. IMO it's not worth serving an `Embedding` module just for that functionality. 

Setting the initialization scale can be done easily with just torch like this:

```python
from torch import nn

emb = nn.Embedding(N, D)
nn.init.normal_(emb.weight, std=init_scale)
```

vinamarora8:
Closes #294. See issue for reasoning.


milosobral (approved):
LGTM

PR #307, opened 2026-06-16
tests: changing pipeline_testing so that they install from a REF

### LINKED ISSUE #276: tests: Pipeline testing should not clone the repo
Our pipeline-testing workflows currently clone/checkout the repo, and then perform the example pipeline preparations. 
This fails to tests whether the brainsets flow would work when someone `pip install`s the package. A recent lack of detection of this failure mode resulted in #273.

I propose that in the pipeline-testing workflows we install the package from a git ref as if it was a user installing it using pip.

milosobral:
Removed the checkout step and updated the project installation command to use the appropriate git reference based on the event type. Given that we no longer checkout, I also simplified the brainsets preparation commands by removing the `uv run` prefix as we now install in the system python directly.

Closes #276

### DIFF
--- a/.github/workflows/pipeline_testing.yml
+++ b/.github/workflows/pipeline_testing.yml
@@ -66,7 +66,6 @@ jobs:
     - name: Download kemp_sleep_edf_2013
       id: kemp_sleep_edf_2013
       continue-on-error: true
-      if: matrix.python-version != '3.9'
       run: brainsets prepare kemp_sleep_edf_2013 -s sleep_cassette_SC4002E0-PSG
 
     # TODO: SSL cert issue with data-proxy.ebrains.eu, re-enable when fixed
--- a/.github/workflows/pipeline_testing.yml
+++ b/.github/workflows/pipeline_testing.yml
@@ -36,7 +36,7 @@ jobs:
           REPO_URL="${{ github.server_url }}/${{ github.repository }}.git"
           REF="${{ github.sha }}"
         fi
-        uv pip install --system "torch_brain[dev] @ git+${REPO_URL}@${REF}"
+        uv pip install --system "torch_brain @ git+${REPO_URL}@${REF}"
 
     - name: Configure brainsets
       run: brainsets config set --raw-dir data/raw --processed-dir data/processed
--- a/.github/workflows/pipeline_testing.yml
+++ b/.github/workflows/pipeline_testing.yml
@@ -25,6 +25,9 @@ jobs:
         python-version: "3.10"
 
     - name: Install the project from git ref
+      # For PRs, we use the head repo URL and SHA to support fork PRs where the
+      # commit doesn't exist in the base repo. For pushes, we use the base repo
+      # URL and the pushed SHA directly.
       run: |
         if [ "${{ github.event_name }}" = "pull_request" ]; then
           REPO_URL="${{ github.event.pull_request.head.repo.clone_url }}"
--- a/.github/workflows/pipeline_testing.yml
+++ b/.github/workflows/pipeline_testing.yml
@@ -14,8 +14,6 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v4
-
     - name: Install uv
       uses: astral-sh/setup-uv@v5
       with:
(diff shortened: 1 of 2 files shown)


.github/workflows/pipeline_testing.yml
```diff
@@ -26,69 +24,76 @@ jobs:
...
       id: flint_slutzky_accurate_2012
       continue-on-error: true
-      run: uv run brainsets prepare flint_slutzky_accurate_2012 -s flint_2012_e1
+      run: brainsets prepare flint_slutzky_accurate_2012 -s flint_2012_e1
 
     - name: Download kemp_sleep_edf_2013
       id: kemp_sleep_edf_2013
       continue-on-error: true
       if: matrix.python-version != '3.9'
-      # scikit-learn>=1.7.2 is not compatible with python3.9
```
AlexandreAndr, line 64: can we keep this comment

.github/workflows/pipeline_testing.yml
```diff
@@ -26,69 +24,76 @@ jobs:
       with:
         python-version: "3.10"
 
-    - name: Install the project
-      run: uv sync --all-extras
+    - name: Install the project from git ref
+      run: |
+        if [ "${{ github.event_name }}" = "pull_request" ]; then
```
AlexandreAndr, line 32: To be sure the if is to ensure when someone is on a PR we take the PR ref and otherwise when we merge on the main branch it take main branch? 
What happens on people commiting on the branch of the PR is it failing on the if rule?

milosobral: Not sure I understand your second question but I added a comment clarifying the if statement

.github/workflows/pipeline_testing.yml
milosobral: I think it's actually not needed at all anymore (including the if statement) since we changed to testing with 3.10

AlexandreAndr: Sure that is even better!

.github/workflows/pipeline_testing.yml
AlexandreAndr: super clear now thanks!

AlexandreAndr (dismissed):
LGTM!

.github/workflows/pipeline_testing.yml
```diff
@@ -26,69 +24,78 @@ jobs:
...
+      # URL and the pushed SHA directly.
+      run: |
+        if [ "${{ github.event_name }}" = "pull_request" ]; then
+          REPO_URL="${{ github.event.pull_request.head.repo.clone_url }}"
+          REF="${{ github.event.pull_request.head.sha }}"
+        else
+          REPO_URL="${{ github.server_url }}/${{ github.repository }}.git"
+          REF="${{ github.sha }}"
+        fi
+        uv pip install --system "torch_brain[dev] @ git+${REPO_URL}@${REF}"
```
vinamarora8, line 40: Maybe this? Or was there a reason to install the `[dev]` version?
```suggestion
        fi
        uv pip install --system "torch_brain @ git+${REPO_URL}@${REF}"

```

vinamarora8 (commented):
One question!

.github/workflows/pipeline_testing.yml
milosobral: There is no specific reason other than to match the existing `--all-extras` but it's true that if we want to test the more general case, users won't be installing dev dependencies so we should probably remove it.

AlexandreAndr: Do we need pytest or things or some dependencies from dev? If pytest is only on dev and maybe we need dev for the pipeline but not sure on this

milosobral: No pipeline-testing just runs basic brainset commands that shouldn't require anything more than just the basic `torch_brain` dependencies. To make sure we actually catch errors, I think it's actually beneficial to not have any extra dependencies that might hide some problems

vinamarora8 (approved):
Thanks for making the change!! LGTM

.github/workflows/pipeline_testing.yml
AlexandreAndr: perfect! LGTM

PR #308, opened 2026-06-16
[brainsets] Add all HBN releases

oumgharbi:
New pipeline `shirazi_hbn` that prepares the [Healthy Brain Network (HBN)](https://openneuro.org/datasets/ds005505/versions/1.0.1) EEG dataset across all 11 OpenNeuro releases in one place. It reuses the OpenNeuroPipeline and groups all 11 datasets into one via minimal overrides.

### Features:

- One pipeline.py for all 11 releases, providing the same channels and processing for all releases.
- Release selection via argument : `--release`
   - `brainsets prepare shirazi_hbn` → prepare all 11 releases
   - `brainsets prepare shirazi_hbn --release 3` → prepare only release 3 (ds005507)
- Per-release layout: creates subdirs for each release for raw and processed data
   - Raw: `.../shirazi_hbn/R1/`, `R2/`, …, `R11/` (BIDS + dataset_description.json per release).
   - Processed: `.../shirazi_hbn_r1/*.h5`, `.../shirazi_hbn_r2/*.h5`, etc. (to keep same structure accessing different brainsets)
- Error handling (HBN pipeline only)
    - Catch invalid `--release`
    - When using `--release` and `--single` simultaneously, error raised if the selected recording does not belong to the selected release.
- Add subject metadata provided in the original dataset.

### Note: 
- `shirazi_hbn --release 1` replaces `shirazi_hbnr1_ds005505` pipeline
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -


torch_brain/datasets/__init__.py
```diff
@@ -113,7 +113,7 @@
                 "ChurchlandShenoyNeural2012",
                 "OdohertySabesNonhuman2017",
                 "VollanMoserAlternating2025",
-                "ShiraziHBNR1DS005505",
+                "ShiraziHBN",
             ],
```
vinamarora8, line 113: I think this should go under the EEG section?

torch_brain/pipeline/brainsets-pipelines/shirazi_hbn/pipeline.py
```diff
@@ -0,0 +1,135 @@
...
+    5: "ds005509",
+    6: "ds005510",
+    7: "ds005511",
+    8: "ds005512",
+    9: "ds005514",
+    10: "ds005515",
+    11: "ds005516",
+}
+
+TYPE_CHANNELS_REMAPPING = {"EEG": [f"E{i}" for i in range(1, 129)] + ["Cz"]}
```
vinamarora8, line 40: Would be great to have a comment explaining the source of these numbers would be (i.e. where should one go to look up if I want to verify these)

torch_brain/pipeline/brainsets-pipelines/shirazi_hbn/pipeline.py
```diff
@@ -0,0 +1,135 @@
...
+
+parser = ArgumentParser(parents=[base_openneuro_parser], add_help=False)
+parser.add_argument(
+    "--release",
+    type=int,
+    default=None,
+    help="Prepare only this release (1-11). Omit to prepare all releases.",
+)
```
vinamarora8, line 48: Perhaps it would be better to provide the "id" of the release instead of index? Like `--release ds005505`, then I think a user would have a better time relating it to the ID on OpenNeuro?

torch_brain/pipeline/brainsets-pipelines/shirazi_hbn/pipeline.py
```diff
@@ -0,0 +1,135 @@
...
+class Pipeline(OpenNeuroPipeline):
+    modality = "eeg"
+    brainset_id = "shirazi_hbn"
+    dataset_id = "ds005505"
+    description = (
+        "Healthy Brain Network (HBN) EEG dataset combining all 11 data releases."
+        "Contains recordings from participants performing various passive"
+        "and active tasks including resting state, movie watching, "
+        "and cognitive tasks."
+    )
```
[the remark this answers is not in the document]
vinamarora8: Easy fix, can u do it pls

torch_brain/pipeline/brainsets-pipelines/shirazi_hbn/pipeline.py
```diff
@@ -0,0 +1,135 @@
...
+
+        self.dataset_id = manifest_item.release_dataset_id
+        self.brainset_id = f"shirazi_hbn_r{release_id}"
+        self.raw_dir = original_raw_dir / f"R{release_id}"
+
+        target_processed_dir = original_processed_dir.with_name(
+            f"{original_processed_dir.name}_r{release_id}"
+        )
+        target_processed_dir.mkdir(parents=True, exist_ok=True)
+        self.processed_dir = target_processed_dir
```
vinamarora8: Woah, overriding `_run_item()` is not a good idea. It will keep evolving, and was not meant to be overridden by user-code (pipelines are "user-code" in this regard). Can you achieve what you want without doing this? Or would be good to list what is the minimum extra functionality needed from `BrainsetPipeline` to support your use case.

vinamarora8 (commented):
Amazing that you're adding (effectively) 11 brainsets! My main concern right now is with overriding `_run_item()` method. Can you address my question below please?

torch_brain/pipeline/brainsets-pipelines/shirazi_hbn/pipeline.py
vinamarora8: I am assuming that `ds005505` is a more recognizable id for this than the index, and I could be wrong in that assumption

milosobral: @oumgharbi can correct me if I'm wrong but I think the index makes sense in this case. All sources in the dataset's own documentation refer to the release indexes (example [here](https://fcon_1000.projects.nitrc.org/indi/cmi_healthy_brain_network/MRI_EEG.html)) and OpenNeuro is a secondary thing where people upload their datasets to make them more easily available. I think it would be more useful for someone wanting to use this brainset to be able to directly refer to the original documentation rather than having to do a translation from OpenNeuro ID to release index every time

oumgharbi: We would rather follow the original index following the [dataset paper](https://doi.org/10.1101/2024.10.03.615261) as Milo said. The OpenNeuro IDs are specific to the OpenNeuro repository and are included here solely to enable compatibility with the `OpenNeuroPipeline`.

torch_brain/pipeline/brainsets-pipelines/shirazi_hbn/pipeline.py
oumgharbi: Oups, that's fair! 😅  The idea was to treat each release as a separate brainset, all while using the same `brainset prepare shirazi_hbn`, by looping over the call, and setting release-specific `dataset_id`, `raw_dir` and `processed_dir`. I found that `_run_item()` was the spot that wraps these configs for both `download()` and `process()` down the line.

What I could do alternatively is (maybe less clean but safer) :
- Option 1 : override `download()` and `process()` to set these release-specifc configs, then continue with their `super()`
- Option 2: simply make 11 separate brainset piplines, one for each release. (`shirazi_hbnr1_ds005505`, `shirazi_hbnr2_ds005506`, etc.)

Otherwise, if you think this is a strong use case that may come back, I would suggest to add a public hook to setup such configs like 
```
def setup_item(self, manifest_item):
   "Hook called before each download() and process()"
    # self.dataset_id = manifest_item.dataset_id
    # self.brainset_id = ...
    # self.raw_dir = ...
    # self.processed_dir = ...
```
Then call that in the private `_run_item()`.
This is what I thought of, please let me know if you have any other suggestions! 🙂

torch_brain/pipeline/brainsets-pipelines/shirazi_hbn/pipeline.py
vinamarora8: Okay, indices sound the correct way to go to me! Thanks for entertaining my comment :)

torch_brain/pipeline/brainsets-pipelines/shirazi_hbn/pipeline.py
vinamarora8: Hmm I see there's a conceptual clash here. You're viewing this one brainset as itself consisting of 11 different brainsets, where each brainset gets it's own directory. However, `BrainsetPipeline` is meant to be a pipeline for only a single brainset and should result in a single directory of data being produced.

Is there any way to operate on these 11 datasets as if they were a _single_ brainset, instead of 11 separate ones? Or will that be completely misaligned with the philosophy of the underlying data?

Of course, like you also suggested, we can have 11 separate pipelines. I would like to attempt thinking of some more elegant solution that tha, without jumping to the idea that `BrainsetPipeline` should start supporting processing of more than one brainset at a time.

vinamarora8: The fact that you have a single `Dataset` for all 11, makes me think that you do actually want to treat it as a single brainset.

oumgharbi: The HBN is supposedly one big dataset, all recordings followed the same protocols, and the same inclusion/exclusion criteria were applied, so any processing should be applied the same to all of them. They were published in 11 different releases I guess because of some technical limitations. 
Ideally, I would prefer to treat it as one brainset, but also keep the release-level granularity. maybe we can simply add this as a recording/subject metadata, and store all in a single directory.

oumgharbi: Also, as you said, I don't currently see the multi-dataset brainset as use case strong enough to add explicit support for it in `BrainsetPipeline`. Maybe if this is recurrent enough, we can revisit the idea.

vinamarora8: Ok great! Looks like we want to treat all of HBN combined as a single brainset!

oumgharbi: Ok then, I will make necessary changes to still keep the `--release` option, but without overriding `_run_item()`. All will be in the same directory, maybe with an added prefix for the processed files like `HBN_R1_*.h5`, and the metadata also.

torch_brain/pipeline/brainsets-pipelines/shirazi_hbn/pipeline.py
```diff
@@ -0,0 +1,184 @@
...
+                        dataset_id,
+                        exc,
+                    )
+                    skipped_releases.append((release_id, dataset_id))
+                    continue
+                manifest["release_id"] = release_id
+                manifest["release_dataset_id"] = dataset_id
+                all_manifests.append(manifest)
+        finally:
+            cls.dataset_id = original_dataset_id
```
milosobral, line 100: I think this code is a bit overly complex and convoluted, with the nested try-catch + the mutation of cls.dataset_id in a loop. Maybe you can modify the get_manifest() to use an optional argument for the dataset_id and find a way to avoid the try-catches?

torch_brain/pipeline/brainsets-pipelines/shirazi_hbn/pipeline.py
```diff
@@ -0,0 +1,184 @@
...
+# ///
+
+import logging
+from argparse import ArgumentParser
+from pathlib import Path
+
+import h5py
+import pandas as pd
+
+from torch_brain.data import serialize_fn_map
```
milosobral: I think this is deprecated, you can probably remove it

torch_brain/pipeline/brainsets-pipelines/shirazi_hbn/pipeline.py
```diff
@@ -0,0 +1,184 @@
...
+                value = row.get(field, None)
+                return None if pd.isna(value) else value
+
+            data.subject.ehq_total = _value("ehq_total")
+            data.subject.commercial_use = _value("commercial_use")
+            data.subject.full_pheno = _value("full_pheno")
+            data.subject.p_factor = _value("p_factor")
+            data.subject.attention = _value("attention")
+            data.subject.internalizing = _value("internalizing")
+            data.subject.externalizing = _value("externalizing")
```
milosobral: Pretty minor but this could probably be a loop

torch_brain/pipeline/brainsets-pipelines/shirazi_hbn/pipeline.py
```diff
@@ -0,0 +1,184 @@
...
+            self._base_raw_dir = self.raw_dir
+        return self._base_raw_dir / f"R{int(release_id)}"
+
+    def _get_participants_data(self, release_id) -> pd.DataFrame | None:
+        # Cache once per release so process() does not re-read the TSV for every recording.
+        if not hasattr(self, "_participants_by_release"):
+            self._participants_by_release = {}
+        release_id = int(release_id)
+        if release_id not in self._participants_by_release:
+            participants_path = self.raw_dir / "participants.tsv"
```
milosobral: I understand that download and process use this relatively fragile mechanism where you have to set `self.raw_dir` to the right path for each release because you want to reuse `super.download` and `super.process`, but maybe the function could take an explicit `raw_dir` parameter to avoid the same flaky dependency here

milosobral (commented):
Added a few comments! 
Overall looks good, but I don't see anything related to the actual tasks of the dataset and the different splits that might come along with that like the different visual stimuli or memory tasks. Am I missing something?

torch_brain/pipeline/brainsets-pipelines/shirazi_hbn/pipeline.py
oumgharbi: I see your concern about  mutating `cls.dataset_id`, but it's there since the base `get_manifest`  reads it implicitly. Looping over many datasets isn't an explicit use case for now. 
The second try catch is also there following the behavior of base `get_manifest` when a dataset has no recordings. 
I agree that this is not a very elegant workaround. The cleanest fix would be to modify the base `get_manifest`, adding an optional explicit dataset_id and maybe a flag to allow empty datasets. However, I am concerned to further delay adding this brainset if we open another dedicated PR. Would it be possible to work with this functional-but-not-amazing code for now, and we can revisit the cleaner fix in another PR, both for the base method and for this pipeline?  What do you think?

torch_brain/pipeline/brainsets-pipelines/shirazi_hbn/pipeline.py
oumgharbi: Yess I agree. I will actually capture the path through the ongoing `_release_raw_dir(release_id)`. This should keep a single source of truth. I would avoid setting an explicit parameter just to avoid possible inconsistency between `release_id` and `raw_dir` depending on when you call this function.

oumgharbi:
Thank you @milosobral for your feedback! I addressed you comments to the best I can.
I agree that this PR has multiple workarounds, but it is mainly because the base brainset pipeline is not designed to handle multiple datasets in a loop. If we find that is it a good use case, I would be worth it to make some tweaks in the base openneuro pipeline. 

And you are right about the tasks and splits, I did not add any task-aware data handling or annotation in this PR. i only focused on extending the [shirazi_hbnr1_ds005505](https://github.com/neuro-galaxy/torch_brain/tree/main/torch_brain/pipeline/brainsets-pipelines/shirazi_hbnr1_ds005505) to include all HBN releases. 
For now, HBN still haven't released the eye tracking data yet. We could definitely update when they do.

PR #313, opened 2026-06-19
feat: Add utils for OpenNeuro participants metadata

oumgharbi:
### Summary 

Adds OpenNeuro utils for `participants.tsv` containing participants metadata in OpenNeuro datasets
- `download_participants_tsv` to locally download the file
- `parse_participants_tsv` for standard parsing of the file
- tests for both utils

This will be useful for brainsets using `openneuro.py` utils.

### DIFF
--- a/tests/utils/test_openneuro.py
+++ b/tests/utils/test_openneuro.py
@@ -1,5 +1,6 @@
 """Unit tests for OpenNeuro S3 utility functions."""
 
+from io import BytesIO
 from pathlib import Path
 from unittest.mock import MagicMock, patch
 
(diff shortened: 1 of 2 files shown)


milosobral (commented):
I think this functionality already exists in `utils/openneuro.py` in the function `fetch_participants_tsv`.
I actually noticed that we also have a parse_participants_tsv in utils/bids which is a bit redundant so we might want to unify things (have the former use the latter probably)
Let me know if I'm missing anything and this is actually different

oumgharbi:
@milosobral  `fetch_participants_tsv` will fetch the file from S3 and load it only in cache. It's what i am currently using in brainsets pipelines, but this will fetch the file for every manifest item. Wanted to optimize by downloading a local version, same as it is done for `dataset_description.json`. 

Also, you are right about the `load_participants_tsv`, that is basically a duplication of `load_participants_tsv`. Thanks for catching that!

milosobral:
@oumgharbi I think this is starting to be a lot of the same code (or very similar) being repeated in a bunch of different places. Maybe we can try to cleanup the API between openneuro utils, bids utils and s3 utils?

oumgharbi:
True, the utils could use a cleanup. For example , a lot of the code between `download_dataset_description` and `download_participants_tsv` is redundant, and could be merged into one util, with an added argument.

oumgharbi:
More generic changes are addressed in #318.

PR #316, opened 2026-06-22
feat: Hide serialize_fn_map; add get_default_serialize_fn_map()

### LINKED ISSUE #286: Usage of `serialize_fn_map` can be hidden/simplified

vinamarora8:
Previously, `torch_brain.data.serialize_fn_map` was a public constant that callers had to pass into `Data.to_hdf5`. This PR hides it behind a private default and makes `Data.to_hdf5` apply it automatically, so users no longer need to interact with `serialize_fn_map` unless they want custom serialization.

The default map now lives in the private constant `torch_brain.data.serialization._DEFAULT_SERIALIZE_FN_MAP`. For users who *do* want to extend it, we expose a public getter `get_default_serialize_fn_map()` which returns a fresh copy each call so mutating the result can never corrupt the global default (a very painful Python footgun):

```python
from torch_brain.data import get_default_serialize_fn_map

serialize_fn_map = get_default_serialize_fn_map()
serialize_fn_map[MyType] = my_serialize_fn
data.to_hdf5(file, serialize_fn_map=serialize_fn_map)
```

**Changes**

- `Data.to_hdf5` uses `_DEFAULT_SERIALIZE_FN_MAP` as its default (`serialize_fn_map=None`).
- Public extension point: `get_default_serialize_fn_map()` returns a safe-to-mutate copy of the defaults.
- Backwards compatibility: `torch_brain.data.serialize_fn_map` is still importable but emits `DeprecationWarning`.
- Updated Pipelines that passed `serialize_fn_map` explicitly, and previous tests.
- Tests added.
- Migration doc updated with info about this change.

Closes #286

### DIFF
--- a/docs/source/guides/pipeline/create_pipeline.rst
+++ b/docs/source/guides/pipeline/create_pipeline.rst
@@ -209,7 +209,7 @@ it into processed :obj:`torch_brain.data.Data` object(s), and stores these insid
         # save data to disk
         self.update_status("Storing")
         with h5py.File(output_file_path, "w") as file:
-            data.to_hdf5(file, serialize_fn_map=serialize_fn_map)
+            data.to_hdf5(file)
 
 
 Most of the logic for implementing the |process| method will follow the tutorial
--- a/docs/source/guides/pipeline/openneuro_pipeline.rst
+++ b/docs/source/guides/pipeline/openneuro_pipeline.rst
@@ -376,9 +376,8 @@ Add custom processing beyond the default:
         
         # Save the result
         import h5py
-        from brainsets import serialize_fn_map
         with h5py.File(store_path, "w") as file:
-            data.to_hdf5(file, serialize_fn_map=serialize_fn_map)
+            data.to_hdf5(file)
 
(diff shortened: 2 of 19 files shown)


divyansha1115 (dismissed):
lgtm! 

One small comment, looks like the docs haven’t been updated. Not sure if you’re planning to do that later.

vinamarora8:
@divyansha1115 thanks for pointing that out! I have removed usage of `serialize_fn_map` from the guides.

divyansha1115 (approved):
lgtm

AlexandreAndr (approved):
Really cool feature, most people probably did not even understood why this serialize was needed!

PR #318, opened 2026-06-22
feat: Refactor and clean S3 / BIDS / OpenNeuro utilities

oumgharbi:
### Summary
Cleans up redundant utilities across `torch_brain/utils/{s3,openneuro,bids}.py`, generalizes metadata downloads, and tightens GraphQL helpers.

#### Changes
- `s3.py`: add generic `download_object` for generic single-object download
- `bids.py`: extract shared `_parse_participants_tsv`
    - reused by `bids.load_participants_tsv` and `openneuro.fetch_participants_tsv`.
- `openneuro.py`: add generic `download_meta` (built on `s3.download_object`) 
    - keep thin wrappers for `download_dataset_description` and `download_participants_tsv`
    - add `redownload` flag to `download_dataset_description`
    - Clean up GraphQL code and use shared helper `_fetch_dataset_graphql`

- `pipeline/openneuro.py`: update the call for `download_dataset_description`
- Tests updated/added accordingly.

#### Notes
- No breaking changes to public function names.
- Download API generalized and de-duplicated.


torch_brain/utils/openneuro.py
milosobral, line 103: I think as part of this refactor, it would be great if this function could also be completely agnostic to the underlying S3 code. This means that it would no longer manually interact with the s3client as all of this would be handled on the s3.py side. Maybe we can event think about always downloading the file since we have a function to do that?

milosobral (commented):
Added a small comment, otherwise looks good!

torch_brain/utils/openneuro.py
oumgharbi: Actually as a cleaner refactor, i would push to entirely remove this function, and replace it with download option. I don't see a strong reason to keep just the in-memory loading of metadata. 
I agree with you on fully removing any direct interaction with the s3 client in `openneuro.py`. I will add a helper to get objects from s3, and then use it to download object to disk.

oumgharbi: On a second thought, might keep `fetch_participants_tsv` with the in-memory option to keep discoverability of manifest items (subject age/sex at manifest). And make it also s3-agnostic by using the same helper.

torch_brain/utils/openneuro.py
```diff
@@ -230,52 +192,140 @@ def download_recording(s3_url: str, target_dir: Path) -> list[Path]:
...
+    return result
+
+
+def download_dataset_description(
+    dataset_id: str,
+    target_dir: Path,
+    *,
+    redownload: bool = False,
+) -> Path:
     """Download dataset_description.json from OpenNeuro S3.
```
vinamarora8, line 244: I'm thinking this wrapper is now too thin, and maybe shouldn't even exist now? `download_meta` is a nice generalization of it which also works for `download_participants_tsv` (same comment for that fn.) Thoughts?

vinamarora8: But I do realize this is a nit and maybe not worth breaking public API.

torch_brain/utils/openneuro.py
vinamarora8: This is resolved, right?

tests/utils/test_openneuro.py
```diff
@@ -322,6 +208,90 @@ def test_delegates_to_download_prefix_from_url(self, mock_download):
...
+        result = fetch_participants_tsv("ds005085")
+
+        mock_get_object_bytes.assert_called_once_with(
+            OPENNEURO_S3_BUCKET, "ds005085/participants.tsv"
+        )
+        assert result is not None
+        assert list(result.index) == ["sub-01"]
+        assert result.loc["sub-01", "age"] == 34
+        assert result.loc["sub-01", "sex"] == "F"
+
```
vinamarora8, line 293: Maybe stupid comment, but seems like the number of tests for `openneuro.fetch_participants_tsv` have reduced from `main`? It that not concerning?

vinamarora8 (commented):
Thanks for the cleanup! Mostly LGTM to me, left some minor comments.

torch_brain/utils/openneuro.py
oumgharbi: I would rather keep the thin wrappers, and not break public API. `download_meta` is there for any meta file, but it would mostly be used for `dataset_description.json` and `participants.tsv`, which should be present in a BIDS-compliant dataset.

tests/utils/test_openneuro.py
oumgharbi: I recreated these test cases testing `s3.get_object_bytes` or `bids._parse_participants_tsv` `test_bids.TestParseParticipantsTsv`. We can also keep the old tests as double checks.

torch_brain/utils/openneuro.py
oumgharbi: Yes, it is resolved for me, just wanted to get your feedback in case you had another perspective.

torch_brain/utils/openneuro.py
```diff
@@ -165,6 +129,9 @@ def fetch_species(dataset_id: str) -> str:
 
     Returns:
         Raw species value returned by OpenNeuro metadata.
+
+    Raises:
```
milosobral: I think the Docstring is out of date and this no longer raises a Runtime Error

torch_brain/utils/openneuro.py
```diff
@@ -230,52 +192,140 @@ def download_recording(s3_url: str, target_dir: Path) -> list[Path]:
...
+        dataset_id,
+        target_dir,
+        "participants.tsv",
+        redownload=redownload,
+    )
 
-    s3_client = get_cached_s3_client()
-    key = f"{dataset_id}/dataset_description.json"
+    if result is None and redownload and target_path.exists() and target_path.is_file():
+        target_path.unlink()
```
milosobral: Shouldn't this belong to the `download_object` function? It should probably be done for objects other than just the participants tsv no?

torch_brain/utils/s3.py
```diff
@@ -127,6 +131,105 @@ def get_object_list(
...
+        RuntimeError: If the download fails for a reason other than the object
+            being absent.
+        ImportError: If boto3/botocore is not installed.
+    """
+    _check_boto_available("download_object")
+    target_path = Path(target_path)
+
+    if target_path.exists() and not redownload:
+        if not target_path.is_file():
+            raise RuntimeError(f"Target path exits and is not a file; {target_path}")
```
milosobral: Small typo here, I think you meant 'exists' 😄

milosobral (commented):
Added a few minor comments but otherwise love how much cleaner this is.

Sorry again that it took me so long to get to this!

torch_brain/utils/openneuro.py
oumgharbi: Indeed, thanks you for catching this! it actually just returns `None` if species not found.

torch_brain/utils/openneuro.py
oumgharbi: Yup, that makes more sense! Updated.

oumgharbi: Now I remembered why I did this 😅 the `participants.tsv` file is a recommended but optional file for BIDS datasets, unlike `dataset_description` which is mandatory, and therefore wouldn't be missing on the s3 (avoiding the whole local stale file vs missing remote file scenario). 
But either way, generalizing to `download_object` is better, because it will handle the mandatory/optional files through the `required` flag.

oumgharbi:
Thank you both @vinamarora8 and @milosobral for your comments. There are few conversations left open, let me know if they are resolved for you or not. Thanks!

milosobral (approved):
LGTM!

vinamarora8 (approved):
LGTM too!

PR #325, opened 2026-07-09
feat: Speed up binning

AlexandreAndr:
## Speed up `bin_spikes` with `np.bincount`

Replaces `np.add.at` in [bin_spikes](torch_brain/utils/binning.py#L61) with a flattened `np.bincount` (much faster scatter-count). Uses `int64` indices to avoid overflow and `minlength` to preserve trailing zero cells.

### DIFF
--- a/torch_brain/utils/binning.py
+++ b/torch_brain/utils/binning.py
@@ -61,11 +61,15 @@ def bin_spikes(
     num_bins = round((end - start) / bin_size)
 
     rate = 1 / bin_size  # avoid precision issues
-    binned_spikes = np.zeros((num_bins, num_units), dtype=dtype)
     # Handle timestamps when the domain start is non-zero
     ts = spikes.timestamps - spikes.domain.start[0]
-    bin_index = np.floor(ts * rate).astype(int)
-    np.add.at(binned_spikes, (bin_index, spikes.unit_index), 1)
+    # int64 cast avoids overflow when num_bins * num_units is large.
+    bin_index = np.floor(ts * rate).astype(np.int64)
+    flat_index = bin_index * num_units + spikes.unit_index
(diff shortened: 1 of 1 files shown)


torch_brain/utils/binning.py
```diff
@@ -61,11 +61,15 @@ def bin_spikes(
...
     # Handle timestamps when the domain start is non-zero
     ts = spikes.timestamps - spikes.domain.start[0]
-    bin_index = np.floor(ts * rate).astype(int)
-    np.add.at(binned_spikes, (bin_index, spikes.unit_index), 1)
+    # int64 cast avoids overflow when num_bins * num_units is large.
+    bin_index = np.floor(ts * rate).astype(np.int64)
+    flat_index = bin_index * num_units + spikes.unit_index
+    # minlength ensures trailing zero-count cells are not truncated before the reshape.
+    falt_binned_spikes = np.bincount(flat_index, minlength=num_bins * num_units)
+    binned_spikes = falt_binned_spikes.reshape(num_bins, num_units).astype(dtype)
```
nandahkrishna: FWIW:
```suggestion
    flat_binned_spikes = np.bincount(flat_index, minlength=num_bins * num_units)
    binned_spikes = flat_binned_spikes.reshape(num_bins, num_units).astype(dtype)
```

AlexandreAndr: thanks

AlexandreAndr:
Should we add a test asserting bit-identical output to the old `np.add.at` path across varied shapes?

vinamarora8 (approved):
This is great! Didn't know bincounts could be that much faster. Good find alex.

I don't think we need extra tests in this PR, since existing tests should cover it. Plus, the logic you've added is simple to verify by eye.

AlexandreAndr:
I discovered when runing binning on collab ^^

PR #326, opened 2026-07-09
chore: Split benchmark script by subpackage; rename data_benchmarks → benchmarks

AlexandreAndr:
## Summary
- Added benchmarks for utils (binning for now).
- Split benchmark CI into two path-scoped workflows: `benchmarks-data.yml` (triggers on `torch_brain/data/**`) and `benchmarks-utils.yml` (triggers on `torch_brain/utils/**`), each posting its own PR comment.
- Added `--suite {data,utils,all}` to `benchmark.py`/`compare.py` so each workflow runs only the relevant suite.
- Fixed stale `scripts/data_benchmarks/**` path filter left over from the `scripts/benchmarks/` rename.

### DIFF
--- a/.github/workflows/benchmarks-data.yml
+++ b/.github/workflows/benchmarks-data.yml
@@ -24,6 +24,7 @@ jobs:
         uses: actions/checkout@v6
         with:
           ref: ${{ github.event.pull_request.head.sha }}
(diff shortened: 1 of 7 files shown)


milosobral (approved):
Love this!

PR #328, opened 2026-07-16
fix: rebind logging handlers to each pipeline item's log file

AlexandreAndr:
## Summary
- Several brainsets pipelines call `logging.basicConfig()` at import time, binding a log handler to whatever `stdout`/`stderr` was current at that moment.
- `_redirect_stdio` only reassigns `sys.stdout`/`sys.stderr` per manifest item afterwards, so those already-bound handlers kept writing to the pre-redirect stream — log records bypassed the per-item `pipeline_logs/<id>.err` file entirely.
- `BrainsetPipeline._run_item` now rebinds logging (`logging.basicConfig(stream=sys.stderr, level=logging.INFO, force=True)`) against the current stream at the start of every item, covering both the parallel-worker path (redirected) and the `--single` path (real stderr). `_redirect_stdio` saves/restores the root logger's handlers and level so this doesn't leak between items.
- Removed the now-redundant `logging.basicConfig(level=logging.INFO)` calls from the 4 pipelines that had them (`odoherty_sabes_nonhuman_2017`, `kemp_sleep_edf_2013`, `allen_visual_coding_ophys_2016`, `neuroprobe_2025`).

## Test plan
- Added `tests/pipeline/test_pipeline_logging.py`: a minimal pipeline whose `download()` logs and `process()` prints; asserts the log record lands in `pipeline_logs/<id>.err`, the print output lands in `pipeline_logs/<id>.out`, and that root logger handlers/level are restored after the item finishes.
- Verified the logging test fails without the fix (message leaks to the real stderr instead of the log file) and passes with it.

### DIFF
--- a/tests/pipeline/test_pipeline_logging.py
+++ b/tests/pipeline/test_pipeline_logging.py
@@ -16,7 +16,8 @@ from torch_brain.pipeline.pipeline import BrainsetPipeline
 
 
 class _LoggingPipeline(BrainsetPipeline):
-    """Minimal pipeline whose download() emits a log record."""
+    """Minimal pipeline whose download() emits a log record and whose
+    process() prints to stdout."""
 
     brainset_id = "logging_test"
 
@@ -29,7 +30,7 @@ class _LoggingPipeline(BrainsetPipeline):
         return None
 
     def process(self, download_output):
-        pass
+        print("hello from process")
 
 
 @pytest.fixture(autouse=True)
@@ -68,6 +69,16 @@ def test_logging_lands_in_per_item_log_file(pipeline, tmp_path):
 
     log_file = tmp_path / "processed" / "pipeline_logs" / "item0.err"
     assert "hello from download" in log_file.read_text()
+
+
+def test_print_output_lands_in_per_item_out_file(pipeline, tmp_path):
+    # print() goes through plain sys.stdout redirection (no logging handler
+    # involved), so this isn't exercising the bug above — it just confirms
+    # _redirect_stdio's other half, the .out file, behaves as expected.
+    pipeline._run_item_on_parallel_worker(pd.Series({"Index": "item0"}))
+
+    log_file = tmp_path / "processed" / "pipeline_logs" / "item0.out"
+    assert "hello from process" in log_file.read_text()
 
 
 def test_restores_logging_handlers_after_item_runs(pipeline, tmp_path):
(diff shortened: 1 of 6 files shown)


torch_brain/pipeline/pipeline.py
```diff
@@ -164,6 +165,7 @@ def update_status(self, status: str):
     def _run_item(self, manifest_item):
         """Run download and process for a manifest item."""
         self._asset_id = manifest_item.Index
+        logging.basicConfig(stream=sys.stderr, level=logging.INFO, force=True)
         output = self.download(manifest_item)
```
vinamarora8, line 169: Do u think this might be better done inside the `_redirect_stdio` context manager? So that all things related to stdio are concentrated in that code block? For e.g., I think you should be able to do it in the `with` section (around line 198)?

torch_brain/pipeline/pipeline.py
```diff
@@ -191,6 +193,9 @@ def _redirect_stdio(log_out_path, log_err_path):
     This is useful when running pipelines in parallel."""
     stdout_prev = sys.stdout
     stderr_prev = sys.stderr
+    root_logger = logging.getLogger()
+    handlers_prev = root_logger.handlers[:]
+    level_prev = root_logger.level
```
vinamarora8, line 202: Not necessary (at all): If there's any reference material that points that only `logger.handlers`, and `logger.level` need to be saved and re-stored; that would be amazing to add as a comment. Right now, it leaves me uncertain that there might be some other attributes of `logger` that we're messing around with and would need restoring. This is NOT REQUIRED, just pointing out if u co-incidentally knew of such a source. :)

tests/pipeline/test_pipeline_logging.py
vinamarora8, line 1: Good tests!

vinamarora8 (commented):
Apologies for the delay in reviewing, Alex! The motivation and implementation mostly look sound to me! Leaving comments for minor things.

tests/pipeline/test_pipeline_logging.py
AlexandreAndr: Thanks 😊 !

torch_brain/pipeline/pipeline.py
AlexandreAndr: I think it would be more elegant the problem is that `_redirect_stdio` is not the only place that call `_run_item`. there is also the case where the pipelined is triggered with -s. In this case we just call `_run_item` without entering `_run_item_on_parallel_worker` but we would still need to call `logging.basicConfig`

torch_brain/pipeline/pipeline.py
AlexandreAndr: I added this ref!
https://docs.python.org/3/library/logging.html#logging.basicConfig

Issue #327, opened 2026-07-11
Option to Delete Raw at the end of Brainset Pipelines


AlexandreAndr:
Most people would not need to keep a version of the raw data or at least it could be nice to have the option to not save it. Especially when the dataset are really big and people have some storage constrains. 
From that reflexion 2 questions come to my mind:
- Should it be a flag that is by default true or false?
- Should we delete them dynamically, like at the end of a file processing or all at the end?

AlexandreAndr:
cc @vinamarora8  @milosobral

milosobral:
To answer both questions: 
- I think it's an option that should by default be False. I think keeping the raw data can be practical if you ever want to reprocess after modifications of the script without having to redownload the data from the source
- Deleting at the end is probably simpler but would be less storage efficient.

AlexandreAndr:
I agree with the point 2, I can try to delete at the end to maximize the efficiency.
For point 1, after think about, the default can be determined based on who we consider as the default user: 
- If it is someone developping the pipeline, I agree we should keep raw to iterate faster 
- If it is a user that just wants to use the dataset and not develop on the pipeline, I am not sure they would need to have the raw data
What do you think?

AlexandreAndr:
Another point for keeping the default as false, it is to not have a change on the current default behavior

milosobral:
I think the "default" user should definitely be # 2, i.e. not a developper of the pipeline, but  I still think that we should keep the default to be to keep the raw data cached. Most libraries that do somewhat similar stuff like pytorch or huggingface datasets seem to keep the raw/compressed versions so you don't have to redownload by default. I also agree with your point that it also mean one less thing in the behavior that changes.

divyansha1115:
I’ve been using Brainsets for large-scale processing and ended up adding a keep_raw flag for this reason.

In practice, I only keep the raw data while debugging or iterating on a pipeline, since it avoids having to download the same files again. Once the pipeline is stable, I prefer the raw data to be deleted immediately after each session has been processed.

For large datasets, keeping all raw files until the end can consume a significant amount of storage. I’ve had pipelines crash because the machine ran out of disk space while raw data was accumulating.

Storage can also be expensive when working with large scale data, so I think retaining raw data should be an explicit choice made by the user rather than the default behavior.

AlexandreAndr:
I agree with both comments! 
I think we can have some difficulty because we have the same interface for user or dev of a brainset pipeline. 
[...]

PR #329, opened 2026-07-17
feat!: Brainset Pipeline Add a `--keep-raw` Flag to Keep Raw Files After Download

AlexandreAndr:


docs/source/cli/commands.rst
```diff
@@ -162,6 +162,12 @@ Download and process a single brainset.
 
 .. rst-class:: cli-option
 
+``--keep-raw``
+
+    Keep each item's raw data after processing (default: delete it).
+
```
vinamarora8, line 168: I would prefer if we instead had a `--delete-raw`. The problem is see with `--keep-raw` is that (potentially) valuable data is downloaded and then deleted from user's disk without them doing anything. With `--delete-raw`, the user "opts-in" into raw deletion. It will also maintain backwards compatibility to current behavior when no extra CLI flags are provided. What do you think?

torch_brain/pipeline/pipeline.py
```diff
@@ -161,12 +167,62 @@ def update_status(self, status: str):
...
+            return [value]
+        if isinstance(value, (list, tuple, set)):
+            paths = []
+            for item in value:
+                paths.extend(BrainsetPipeline._extract_raw_paths(item))
+            return paths
+        path_attr = getattr(value, "path", None)
+        if isinstance(path_attr, Path):
+            return [path_attr]
+        return []
```
vinamarora8, line 216: I think this is quite dangerous, no? If we make a mistake in inferring the download path, or user makes a mistake in what they return from download, we could end up deleting something very important on user's machine. For instance, if someone _accidentally_ returns `/home/<username>` from `download()` (I don't know how, but it doesn't sound impossible to happen), then their home directory is gone.

vinamarora8 (commented):
Interesting feature, Alex! I do see the usefulness of it, since it instantly cleans up raw directory as it processes one manifest item. But it also seems quite unsafe to me at the moment. Left a comment specifically for this below

docs/source/cli/commands.rst
AlexandreAndr: Yeah we had some discussion on the issue #327. I think the first point is valid. I am just worried that the pipeline might not scale well and that we should probably design our pipeline to do not depend on the raw files

AlexandreAndr: We can keep a delete-raw for now and see if we can also optimize in another iteration with compression of the raw data (convert it to zip) at the end of the pipeline

torch_brain/pipeline/pipeline.py
AlexandreAndr: that is a fair point!

PR #330, opened 2026-07-21
feat: Add SessionBatchSampler for grouping samples into per-session batches

AlexandreAndr:
## Summary

Adds `SessionBatchSampler`, a `Sampler` wrapper that groups any inner sampler's `DatasetIndex` output into per-session batches (no batch mixes sessions), with optional batch shuffling and `drop_last` control. Extracted from the NDT multi-session notebook, where it was previously defined inline.

- New: `torch_brain/samplers/session_batch_sampler.py`, registered in `torch_brain/samplers/__init__.py`
- Tests: `tests/test_sampler.py::test_session_batch_sampler` (per-session grouping, `drop_last` True/False, invalid `batch_size`, `len()` consistency)

### DIFF
--- a/torch_brain/samplers/session_batch_sampler.py
+++ b/torch_brain/samplers/session_batch_sampler.py
@@ -49,6 +49,8 @@ class SessionBatchSampler(torch.utils.data.Sampler[list[DatasetIndex]]):
         generator: torch.Generator | None = None,
         drop_last: bool = True,
     ):
+        if isinstance(batch_size, bool) or not isinstance(batch_size, int):
+            raise TypeError("batch_size must be an integer.")
         if batch_size <= 0:
             raise ValueError("batch_size must be positive.")
         self.sampler = sampler
@@ -88,11 +90,16 @@ class SessionBatchSampler(torch.utils.data.Sampler[list[DatasetIndex]]):
         return len(self._batches_cache)
 
     def __iter__(self) -> Iterator[list[DatasetIndex]]:
-        r"""Yields per-session batches of :class:`~torch_brain.datasets.DatasetIndex`."""
+        r"""Returns an iterator over per-session batches of
+        :class:`~torch_brain.datasets.DatasetIndex`."""
         self._prepare_cache()
         batches = self._batches_cache
         self._batches_cache = None  # reset so next epoch rebuilds
+        return self._yield_batches(batches)
 
+    def _yield_batches(
+        self, batches: list[list[DatasetIndex]]
+    ) -> Iterator[list[DatasetIndex]]:
         if self.shuffle_batches and batches:
             for idx in torch.randperm(len(batches), generator=self.generator).tolist():
                 yield batches[idx]
--- a/torch_brain/samplers/session_batch_sampler.py
+++ b/torch_brain/samplers/session_batch_sampler.py
@@ -37,7 +37,11 @@ class SessionBatchSampler(torch.utils.data.Sampler[list[DatasetIndex]]):
         ...     sampling_intervals=sampling_intervals,
         ...     window_length=1.0,
         ... )
-        >>> sampler = SessionBatchSampler(inner_sampler, batch_size=16)
+        >>> batch_sampler = SessionBatchSampler(inner_sampler, batch_size=16)
+        >>> loader = torch.utils.data.DataLoader(
+        ...     dataset=your_dataset,
+        ...     batch_sampler=batch_sampler,
+        ... )
     """
 
     def __init__(
--- a/torch_brain/samplers/session_batch_sampler.py
+++ b/torch_brain/samplers/session_batch_sampler.py
@@ -17,8 +17,8 @@ class SessionBatchSampler(torch.utils.data.Sampler[list[DatasetIndex]]):
         sampler: Inner sampler whose :meth:`__iter__` yields
             :class:`~torch_brain.datasets.DatasetIndex` objects.
         batch_size: Number of samples per batch.
-        shuffle_batches: If ``True``, the final batch order is shuffled.
-            Defaults to ``False``.
+        shuffle_batches: If ``True`` (default), the final batch order is shuffled.
+            If ``False``, batches are yielded in the order they were built.
         generator: Optional RNG used when :obj:`shuffle_batches` is ``True``.
             If ``None`` (default), uses the default global PyTorch generator.
         drop_last: If ``True`` (default), the last incomplete batch for each
@@ -45,7 +45,7 @@ class SessionBatchSampler(torch.utils.data.Sampler[list[DatasetIndex]]):
         sampler: torch.utils.data.Sampler,
         batch_size: int,
         *,
-        shuffle_batches: bool = False,
+        shuffle_batches: bool = True,
         generator: torch.Generator | None = None,
         drop_last: bool = True,
     ):
--- a/torch_brain/samplers/session_batch_sampler.py
+++ b/torch_brain/samplers/session_batch_sampler.py
@@ -10,14 +10,14 @@ class SessionWiseBatchSampler(torch.utils.data.Sampler[list[DatasetIndex]]):
     groups its output into per-session batches.
 
     All indices in a batch share the same session id. The inner sampler controls
-    the order of indices within each session; :obj:`shuffle_batches` controls the
+    the order of indices within each session; :obj:`shuffle` controls the
     order of batches across sessions.
 
     Args:
         sampler: Inner sampler whose :meth:`__iter__` yields
             :class:`~torch_brain.datasets.DatasetIndex` objects.
         batch_size: Number of samples per batch.
-        shuffle_batches: If ``True``, the final batch order is shuffled.
+        shuffle: If ``True``, the final batch order is shuffled.
             If ``False`` (default), batches are yielded in the order they were built.
         cache_batches: If ``True`` (default), batches are built once and reused
             across epochs instead of being rebuilt on every :meth:`__iter__`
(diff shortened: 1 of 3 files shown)


torch_brain/samplers/session_batch_sampler.py
```diff
@@ -0,0 +1,107 @@
...
+        ... )
+        >>> sampler = SessionBatchSampler(inner_sampler, batch_size=16)
+    """
+
+    def __init__(
+        self,
+        sampler: torch.utils.data.Sampler,
+        batch_size: int,
+        *,
+        shuffle_batches: bool = False,
```
divyansha1115: any specific reason to why default shuffle is false? usually default is set to true.

divyansha1115 (commented):
can you update the change log

AlexandreAndr:
Hi, for the changelog I think we are trying to get rid of it by having more informative PR names

torch_brain/samplers/session_batch_sampler.py
AlexandreAndr: That is a good point, we might consider it true by default yes. To keep in mind the sampler that wraps arround it has it's own internal logic for the sampling but I agree that having it by default set to true might be better

AlexandreAndr: just updated it

torch_brain/samplers/session_batch_sampler.py
```diff
@@ -0,0 +1,107 @@
...
+        >>> inner_sampler = RandomFixedWindowSampler(
+        ...     sampling_intervals=sampling_intervals,
+        ...     window_length=1.0,
+        ... )
+        >>> sampler = SessionBatchSampler(inner_sampler, batch_size=16)
+    """
+
+    def __init__(
+        self,
+        sampler: torch.utils.data.Sampler,
```
vinamarora8:
```suggestion
        sampler: torch.utils.data.Sampler[DatasetIndex],
```

I think this will work?

torch_brain/samplers/session_batch_sampler.py
```diff
@@ -0,0 +1,107 @@
...
+
+        >>> sampling_intervals = {
+        ...     "session_1": Interval(0.0, 100.0),
+        ...     "session_2": Interval(0.0, 200.0),
+        ... }
+        >>> inner_sampler = RandomFixedWindowSampler(
+        ...     sampling_intervals=sampling_intervals,
+        ...     window_length=1.0,
+        ... )
+        >>> sampler = SessionBatchSampler(inner_sampler, batch_size=16)
```
vinamarora8: Would be amazing if u also add small snipped for how to use a batch sampler with the torch dataloader!

torch_brain/samplers/session_batch_sampler.py
```diff
@@ -0,0 +1,107 @@
+from collections.abc import Iterator
+
+import torch
+
+from torch_brain.datasets import DatasetIndex
+
+
+class SessionBatchSampler(torch.utils.data.Sampler[list[DatasetIndex]]):
+    r"""Wraps any sampler yielding :class:`~torch_brain.datasets.DatasetIndex` and
```
vinamarora8: Another idea for the name: `SessionWiseBatchSampler`, maybe?

torch_brain/samplers/session_batch_sampler.py
```diff
@@ -0,0 +1,107 @@
...
+        self._prepare_cache()
+        return len(self._batches_cache)
+
+    def __iter__(self) -> Iterator[list[DatasetIndex]]:
+        r"""Returns an iterator over per-session batches of
+        :class:`~torch_brain.datasets.DatasetIndex`."""
+        self._prepare_cache()
+        batches = self._batches_cache
+        self._batches_cache = None  # reset so next epoch rebuilds
+        return self._yield_batches(batches)
```
vinamarora8, line 115:
```suggestion
    def __iter__(self) -> Iterator[list[DatasetIndex]]:
        r"""Returns an iterator over per-session batches of
        :class:`~torch_brain.datasets.DatasetIndex`."""
        batches = self._build_batches()
        return self._yield_batches(batches)
```

Maybe this is simpler? Looks to me that you want to build batches every epoch, so there is no need for a separate `_prepare_cache` etc.. Apologies if I'm wrong!

torch_brain/samplers/session_batch_sampler.py
```diff
@@ -0,0 +1,107 @@
...
+        ... )
+        >>> sampler = SessionBatchSampler(inner_sampler, batch_size=16)
+    """
+
+    def __init__(
+        self,
+        sampler: torch.utils.data.Sampler,
+        batch_size: int,
+        *,
+        shuffle_batches: bool = True,
```
vinamarora8:
```suggestion
        shuffle: bool = False,
```

I think `shuffle_batches` should simply be `shuffle`? And also it should default to `False` to stay consistent with other torch stuff (like how shuffle is False by default in `torch.utils.data.DataLoader`).

torch_brain/samplers/session_batch_sampler.py
vinamarora8:
```suggestion
        >>> inner_sampler = RandomFixedWindowSampler(
        ...     sampling_intervals=sampling_intervals,
        ...     window_length=1.0,
        ... )
        >>> batch_sampler = SessionBatchSampler(inner_sampler, batch_size=16)
        >>> loader = torch.utils.data.DataLoader(
        ...     dataset=your_dataset,
        ...     batch_sampler=batch_sampler,
        ... )
```

vinamarora8 (commented):
A very clean way to make session-wise batches, without making a batch-samplers for each single-sample samplers! Left some comments below

torch_brain/samplers/session_batch_sampler.py
vinamarora8: Hmm.. torch's DataLoader's default value for shuffle is False. @divyansha1115 curious, where have you seen default shuffle being True?

https://docs.pytorch.org/docs/2.13/data.html

torch_brain/samplers/session_batch_sampler.py
```diff
@@ -0,0 +1,107 @@
...
+        r"""Returns the total number of batches across all sessions."""
+        self._prepare_cache()
+        return len(self._batches_cache)
+
+    def __iter__(self) -> Iterator[list[DatasetIndex]]:
+        r"""Returns an iterator over per-session batches of
+        :class:`~torch_brain.datasets.DatasetIndex`."""
+        self._prepare_cache()
+        batches = self._batches_cache
+        self._batches_cache = None  # reset so next epoch rebuilds
```
milosobral: Do we always want to rebuild at every epoch? Should this maybe be an option?

torch_brain/samplers/session_batch_sampler.py
milosobral: I actually think that `shuffle_batches` is slightly better given that it makes it clear that we are shuffling the order of the batches, not the order of the samples inside of each batch

milosobral (commented):
I have a small comment on resetting but otherwise looks great!

torch_brain/samplers/session_batch_sampler.py
AlexandreAndr: yes good point I'll also update another sampler typing to follow this

torch_brain/samplers/session_batch_sampler.py
AlexandreAndr: Done!

torch_brain/samplers/session_batch_sampler.py
AlexandreAndr: It is a good point!

AlexandreAndr: 1) We want to rebuild at each epoch for the random sampler. Otherwise, you would have always the same samples within batches (and batches sequence would be random or not)
2) I added a `cache_batches` args that is `True` by default as shuffle is `False`

torch_brain/samplers/session_batch_sampler.py
AlexandreAndr: Hi, I agree it would be more elegant and I am not a fan of all the subfunctions. 
The need for the `_prepare_cache` comes from the fact that you probably call `__len()__` before `__iter__` and you want the cached from both to be aligned

torch_brain/samplers/session_batch_sampler.py
AlexandreAndr: I am neutral to it, @milosobral @divyansha1115 do you have a preference?

torch_brain/samplers/session_batch_sampler.py
AlexandreAndr: I also like something more descriptive. For the false and true default behavior, I will post something bellow!

torch_brain/samplers/session_batch_sampler.py
divyansha1115: I would have slight preference towards SessionWiseBatchSampler. The initial one sound a little off to me.

torch_brain/samplers/session_batch_sampler.py
divyansha1115: I think shuffle would be better here. Given we call it batch sampler it would be clear that the shuffle is on batches. 
Also all sampler having somewhat similar and consistent arguments is easier to remember and configure.

I also agree on the default being changed to false. I was interpreting the sampler in a different way before, but looking deeper I think False as default is more appropriate.

torch_brain/samplers/session_batch_sampler.py
AlexandreAndr: okay let move to it!

torch_brain/samplers/session_batch_sampler.py
AlexandreAndr: For concistency of the samplers I move it to `false`

AlexandreAndr: and move to `shuffle`

torch_brain/samplers/session_batch_sampler.py
AlexandreAndr: For 2) cc @divyansha1115 @vinamarora8 , What do you think of this arg?
For me the tradeoff is that caching batch can save time for the deterministic samplers but complexify slightly the interface

PR #341, opened 2026-09-11
docs: Fix copybutton for python REPL codeblocks

### LINKED ISSUE #336: Improve ease-of-use of code in the User Guide
**Problem:** The current formatting code in the user guide adds a barrier to easily copying and running the example code. See the [Meet the Data Objects](https://torch-brain.readthedocs.io/en/latest/guides/data/data_objects_intro.html) section, for example.

If the top-right corner copy button is used, chevrons and ellipses are copied alongside the code. They then have to be removed for the code to run
```
>>> import numpy as np
>>> from torch_brain.data import RegularTimeSeries

>>> behavior = RegularTimeSeries(
...     sampling_rate=100.0,  # in Hz
...     hand_vel=np.random.randn(1000, 2),
...     eye_pos=np.random.randn(1000, 2),
...     pupil_size=np.random.randn(1000),
... )
```

Alternatively, if the code is selected and copied, the chevrons and ellipses are ignored. However, printed outputs are still copied and they have to be manually removed to run the code. Relatedly, once the code is copied into a single notebook cell, lines like `behavior` only work as intended (approximately as `print(behavior)`) if they are the last command in the cell.

**Solution:** I'm not sure yet, but I'll look into a few options.

Ideally we want to maintain the current benefit of showing example outputs in a compact and clear manner, as achieved by: 
```
>>> behavior
RegularTimeSeries(
  hand_vel=[1000, 2],
  eye_pos=[1000, 2],
  pupil_size=[1000]
)
```

However, we also want the code to be easily copied and run in a notebook, producing equivalent outputs to what is observed in the guide.

**Relevant questions:** 
1. Do you mind if there are more code blocks that can be copied? That would provide a solution to making lines like `behavior` functional, and still showing the output right below in the guide. The main alternative I see would be to use more elaborate `print` statements (e.g., `print(f"behavior:\n{behavior}")`), and then showing all outputs together each block of code.  
2. Do we want users to be able to download guide pages like this one as notebooks, so they can actively interact with the code?

vinamarora8:
With this change, clicking the copy-button will not copy chevrons and outputs. Test it [here](https://torch-brain--341.org.readthedocs.build/en/341/guides/data/data_objects_intro.html).

Partially fixes #336

The lint and test CI failures are due to unrelated issues.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -95,12 +95,10 @@ myst_enable_extensions = [
 
 nb_execution_mode = "off"
 
-# sphinx-copybutton: for REPL-style blocks (``pycon``, ``console``), copy only the
-# input lines with their prompts stripped, so users don't paste ">>>" or printed
-# output. Blocks with no prompts (plain ``python``/``bash``) are copied in full.
-# Note: do not set ``copybutton_exclude`` to skip ``.gp`` (the prompt spans) -- the
-# prompts must survive text extraction for the regex below to find them.
-copybutton_prompt_text = r">>> |\.\.\. |\$ "
+# Copybutton config to compensate for REPL-style blocks (python, console, ipython).
+# Taken from:
+# https://sphinx-copybutton.readthedocs.io/en/latest/use.html#strip-and-configure-input-prompts-for-code-cells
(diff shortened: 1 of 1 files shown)


colleenjg:
This is great! As I've mentioned in the issue, I think there's still a final element to handle, but this one change will mean that the copied code is able to run as is, so that is great.

colleenjg (approved):

